PR #1, opened 2020-03-18
Renaming tinkerbell to boots across the repo

gauravgahlot:


installers/custom_ipxe/ipxe_script_test.go
```diff
@@ -27,7 +27,7 @@ func TestScript(t *testing.T) {
 		s := ipxe.Script{}
 		s.Echo("Packet.net Baremetal - iPXE boot")
 		s.Set("iface", "eth0").Or("shell")
-		s.Set("tinkerbell", "http://127.0.0.1")
+		s.Set("boots", "http://127.0.0.1")
```
mmlb: this needs to stay tinkerbell to remain backwards compatible with any ipxe scripts out there in the wild

installers/osie/ipxe_script_test.go
```diff
@@ -54,7 +54,7 @@ func TestScript(t *testing.T) {
 			s := ipxe.Script{}
 			s.Echo("Packet.net Baremetal - iPXE boot")
 			s.Set("iface", "eth0").Or("shell")
-			s.Set("tinkerbell", "http://127.0.0.1")
+			s.Set("boots", "http://127.0.0.1")
```
mmlb: I think we can actually get rid of this, since osie doesn't use it, but can leave that for another pr

installers/rancher/main.go
```diff
@@ -30,5 +30,5 @@ func kernelParams(j job.Job, s *ipxe.Script) {
 		s.Args("rancher.network.interfaces.eth2.dhcp=true")
 	}
 
-	s.Args("tinkerbell=${tinkerbell}")
+	s.Args("boots=${boots}")
```
mmlb: we don't manage the rancher script so need to keep kernel args the same as before

job/ipxe.go
```diff
@@ -54,7 +54,7 @@ func (j Job) serveBootScript(w http.ResponseWriter, req *http.Request, name stri
 
 	s := ipxe.NewScript()
 	s.Set("iface", j.InterfaceName(0)).Or("shell")
-	s.Set("tinkerbell", "http://"+env.PublicFQDN)
+	s.Set("boots", "http://"+env.PublicFQDN)
```
mmlb: lets use a different name here, not quite sure what though.

installers/osie/ipxe_script_test.go
gauravgahlot: Shall it be changed back to `tinkerbell`?

job/ipxe.go
gauravgahlot: What should it be changed to then?

mmlb:
@gauravgahlot any changes to ipxe scripts need to be backwards compatible to current settings. That means we need to keep the `set tinkerbell` in the ipxe header bits and we need to keep `tinkerbell=` in the kernel args.

It wouldn't be a bad idea to add new ones but I'm not 100% on having it be `boots`, I'd lke it to be a generic-ish value that isn't tied to a service's name. phone-home is ok though would be a little overloaded with the phone home url that installers also need to know about (and get from metadata). Maybe `events` or something? @nathangoulding suggestions?

nathangoulding:
Agree we should stick with `set tinkerbell` for now in all cases.

@mmlb my 2 cents: I'm not sure that the DHCP server should really be handling events anyways. It _feels_ like that actually belongs in tinkerbell, not boots. In which case, we keep `set tinkerbell` and change the IP/URL to the endpoint of workflow engine, and move the events/phone-home logic there...

nathangoulding:
Moving to `add-drone` PR #2

PR #3, opened 2020-03-18
add go mod

nathangoulding:


.envrc
```diff
@@ -1,2 +1,3 @@
 which nix &>/dev/null && use nix
 unset GOPATH
+export GOPRIVATE=github.com/packethost/cacher
```
mmlb: @nathangoulding we need to open source cacher or remove as a dep, former is probably easiest.

gauravgahlot: To remove cacher as a dep, we need to update boots to use tinkerbell instead of cacher. However, that might affect compatibility with the old data model.

mmlb:
this is going to be taken care of in #1

PR #8, opened 2020-04-03
Support for workflow mode

gauravgahlot:
At present, the default action is `install`. Therefore, for the time being, a user must set the `allow_workflow: true` to bootstrap a worker in workflow mode.

In the near future, this will not be required as every worker will by default boot into a workflow.

### DIFF
--- a/installers/osie/main.go
+++ b/installers/osie/main.go
@@ -3,7 +3,6 @@ package osie
 import (
 	"strings"
 
-	"github.com/packethost/pkg/log"
 	"github.com/tinkerbell/boots/ipxe"
 	"github.com/tinkerbell/boots/job"
 )
@@ -12,8 +11,6 @@ func init() {
 	job.RegisterDefaultInstaller(bootScripts["install"])
 	job.RegisterDistro("alpine", bootScripts["rescue"])
 }
-
-var logger log.Logger
 
 var bootScripts = map[string]func(job.Job, *ipxe.Script){
 	"rescue": func(j job.Job, s *ipxe.Script) {
@@ -39,7 +36,6 @@ var bootScripts = map[string]func(job.Job, *ipxe.Script){
 }
 
 func bootScript(action string, j job.Job, s *ipxe.Script) {
-	logger = j.Logger
 	s.Set("arch", j.Arch())
 	s.Set("parch", j.PArch())
 	s.Set("bootdevmac", j.PrimaryNIC().String())
--- a/installers/osie/main.go
+++ b/installers/osie/main.go
@@ -25,7 +25,11 @@ var bootScripts = map[string]func(job.Job, *ipxe.Script){
 			typ = "deprovisioning.304.1"
 		}
 		s.PhoneHome(typ)
-		s.Set("action", "install")
+		if canWorkflow(j) {
+			s.Set("action", "workflow")
+		} else {
+			s.Set("action", "install")
+		}
 		s.Set("state", j.HardwareState())
 		bootScript("install", j, s)
 	},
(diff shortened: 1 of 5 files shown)


installers/osie/main.go
```diff
@@ -137,3 +155,11 @@ func osieBaseUrl(j job.Job) string {
...
 	return osieURL + "/current"
 }
+
+func workflowBaseURL() string {
+	return mirrorBaseURL + "/workflow"
+}
+
+func canWorkflow(j job.Job) bool {
+	return j.CanWorkflow()
+}
```
mmlb, line 165: I'd rather just call `j.CanWorkflow()` directly

installers/osie/mirror.go
```diff
@@ -42,3 +48,20 @@ func mustBuildOsieURL() *url.URL {
...
+	grpcCertURL = getParam("TINKERBELL_CERT_URL")
+	registryUsername = getParam("REGISTRY_USERNAME")
+	registryPassword = getParam("REGISTRY_PASSWORD")
+	elasticSearchURL = getParam("ELASTIC_SEARCH_URL")
+}
+
+func getParam(key string) string {
+	value := os.Getenv(key)
+	if value == "" {
+		log.Fatal(errors.New("invalid" + key))
```
mmlb, line 64: this should be using `pkg/log` based logger. This way `Fatal` sends to rollbar for us.

installers/osie/mirror.go
```diff
@@ -42,3 +47,20 @@ func mustBuildOsieURL() *url.URL {
...
+	grpcCertURL = getParam("TINKERBELL_CERT_URL")
+	registryUsername = getParam("REGISTRY_USERNAME")
+	registryPassword = getParam("REGISTRY_PASSWORD")
+	elasticSearchURL = getParam("ELASTIC_SEARCH_URL")
+}
+
+func getParam(key string) string {
+	value := os.Getenv(key)
+	if value == "" {
+		logger.Fatal(errors.New("invalid" + key))
```
mmlb, line 63: this should be `installers.Logger("osie").Fatai(...)` instead.

installers/osie/main.go
```diff
@@ -25,13 +28,18 @@ var bootScripts = map[string]func(job.Job, *ipxe.Script){
...
+		} else {
+			s.Set("action", "install")
+		}
 		s.Set("state", j.HardwareState())
 		bootScript("install", j, s)
 	},
 }
 
 func bootScript(action string, j job.Job, s *ipxe.Script) {
+	logger = j.Logger
```
mmlb, line 42: This is just asking for race condition bugs.

installers/osie/main.go
```diff
@@ -12,6 +13,8 @@ func init() {
 	job.RegisterDistro("alpine", bootScripts["rescue"])
 }
 
+var logger log.Logger
```
mmlb, line 16: better to use the `installers.Logger()` where needed in the init function than in a global var.

installers/osie/mirror.go
```diff
@@ -42,3 +48,20 @@ func mustBuildOsieURL() *url.URL {
...
+	grpcCertURL = getParam("TINKERBELL_CERT_URL")
+	registryUsername = getParam("REGISTRY_USERNAME")
+	registryPassword = getParam("REGISTRY_PASSWORD")
+	elasticSearchURL = getParam("ELASTIC_SEARCH_URL")
+}
+
+func getParam(key string) string {
+	value := os.Getenv(key)
+	if value == "" {
+		installers.Logger("osie").Fatal(errors.New("invalid" + key))
```
mmlb, line 64:
```suggestion
		installers.Logger("osie").With("key", key).Fatal(errors.New("invalid key"))
```

PR #15, opened 2020-04-06
Updated Readme 

gauravgahlot:
Updated [README.md](https://github.com/tinkerbell/boots/blob/master/README.md) to have clear steps for a new user. 

Signed-off-by: Gaurav Gahlot <gaurav.gahlot19@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,11 @@ This services handles DHCP, PXE, tftp, and iPXE for provisions.
 First, you need to make sure you have [git-lfs](https://github.com/git-lfs/git-lfs/wiki/Installation) installed:
 
 ```
+# install "git-lfs" package for your OS, i.e.:
+curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
+apt install git-lfs
+
+# then run these two commands:
 git lfs install
 git lfs pull
 ```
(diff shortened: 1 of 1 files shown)


nathangoulding:
Also apt, not apt-get and we should provide separate instructions for rpm-based systems.

README.md
```diff
@@ -8,6 +8,11 @@ This services handles DHCP, PXE, tftp, and iPXE for provisions.
 First, you need to make sure you have [git-lfs](https://github.com/git-lfs/git-lfs/wiki/Installation) installed:
 
 ```
+# install "git-lfs" package for your OS, i.e.:
+curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
+apt install git-lfs
```
mmlb, line 13: these should be comments too.

README.md
```diff
@@ -8,6 +8,11 @@ This services handles DHCP, PXE, tftp, and iPXE for provisions.
 First, you need to make sure you have [git-lfs](https://github.com/git-lfs/git-lfs/wiki/Installation) installed:
 
 ```
+# install "git-lfs" package for your OS, i.e.:
+curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
```
nathangoulding, line 12: Suggest not silently curl.

gauravgahlot:
> Also apt, not apt-get and we should provide separate instructions for rpm-based systems.

@nathangoulding I will be like duplicating steps from [here](https://github.com/git-lfs/git-lfs/wiki/Installation#rhelcentos). We already have a link to a detailed instruction set.

README.md
vielmetti: Strongly suggest not to do `curl | sudo bash` for external dependencies, if any other option is available.

PR #21, opened 2020-04-14
Serve DHCP packets in pooled goroutines

mmlb:
This PR improves the scalability of our dhcp handling. This was hit when trying to do some Open19 tests/verification. This has happened at other times too in our history where dhcp packets queue up for too long causing DHCP & provisions to fail. Those were handled via OUI blacklisting. We will now actually process multiple DHCP packets concurrently.

Along the same lines, we also want to increase the number of idle connections to api to make better use of persistent connections and thus re-use.

Both numbers are tweakable, but the chosen defaults seem pretty good.


nathangoulding:
I'd like remove the required-ness of the API tokens since those are very Packet specific, but if that is a bigger endeavor I'm happy to tackle in a separate PR.

mmlb:
> I'd like remove the required-ness of the API tokens since those are very Packet specific, but if that is a bigger endeavor I'm happy to tackle in a separate PR.

Yep more appropriate for a separate PR.

PR #22, opened 2020-04-22
Add prometheus metrics

mikemrm:
This adds some basic prometheus metrics to give a better insight into what all boots is doing.

### DIFF
--- a/dhcp.go
+++ b/dhcp.go
@@ -58,7 +58,7 @@ func (d dhcpHandler) serveDHCP(w dhcp4.ReplyWriter, req *dhcp4.Packet) {
 		return
 	}
 
-	metrics.DHCPTotal.WithLabelValues(req.GetMessageType().String(), gi.String()).Inc()
+	metrics.DHCPTotal.WithLabelValues("recv", req.GetMessageType().String(), gi.String()).Inc()
 	labels := prometheus.Labels{"from": "dhcp", "op": req.GetMessageType().String()}
 	metrics.JobsTotal.With(labels).Inc()
 	metrics.JobsInProgress.With(labels).Inc()
@@ -80,7 +80,7 @@ func (d dhcpHandler) serveDHCP(w dhcp4.ReplyWriter, req *dhcp4.Packet) {
 	}
 	go func() {
 		if j.ServeDHCP(w, req) {
-			metrics.DHCPTotal.WithLabelValues("reply", gi.String()).Inc()
+			metrics.DHCPTotal.WithLabelValues("send", "DHCPOFFER", gi.String()).Inc()
 		}
 		metrics.JobsInProgress.With(labels).Dec()
 		timer.ObserveDuration()
--- a/dhcp.go
+++ b/dhcp.go
@@ -9,8 +9,10 @@ import (
 	dhcp4 "github.com/packethost/dhcp4-go"
 	"github.com/packethost/pkg/env"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/tinkerbell/boots/conf"
 	"github.com/tinkerbell/boots/job"
+	"github.com/tinkerbell/boots/metrics"
 )
 
 var listenAddr = conf.BOOTPBind
@@ -56,6 +58,12 @@ func (d dhcpHandler) serveDHCP(w dhcp4.ReplyWriter, req *dhcp4.Packet) {
 		return
 	}
 
+	metrics.DHCPTotal.WithLabelValues(req.GetMessageType().String(), gi.String()).Inc()
+	labels := prometheus.Labels{"from": "dhcp", "op": req.GetMessageType().String()}
+	metrics.JobsTotal.With(labels).Inc()
+	metrics.JobsInProgress.With(labels).Inc()
+	timer := prometheus.NewTimer(metrics.JobDuration.With(labels))
+
 	circuitID, err := getCircuitID(req)
 	if err != nil {
 		mainlog.With("mac", mac, "err", err).Info("error parsing option82")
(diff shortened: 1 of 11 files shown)


dhcp.go
```diff
@@ -56,6 +58,12 @@ func (d dhcpHandler) serveDHCP(w dhcp4.ReplyWriter, req *dhcp4.Packet) {
 		return
 	}
 
+	metrics.DHCPTotal.WithLabelValues(req.GetMessageType().String(), gi.String()).Inc()
+	labels := prometheus.Labels{"from": "dhcp", "op": req.GetMessageType().String()}
+	metrics.JobsTotal.With(labels).Inc()
+	metrics.JobsInProgress.With(labels).Inc()
```
mmlb, line 64: why not a `defer`?

dhcp.go
```diff
@@ -56,6 +58,12 @@ func (d dhcpHandler) serveDHCP(w dhcp4.ReplyWriter, req *dhcp4.Packet) {
...
+	timer := prometheus.NewTimer(metrics.JobDuration.With(labels))
```
mmlb, line 65: ditto re: defer

dhcp.go
mikemrm: because since j.ServeDHCP is being handled in a go routine, i don't want it to mark the job as complete until that has been served. So I wait for it to complete first before cleaning up

dhcp.go
```diff
@@ -66,9 +74,17 @@ func (d dhcpHandler) serveDHCP(w dhcp4.ReplyWriter, req *dhcp4.Packet) {
...
 	if err != nil {
 		mainlog.With("type", req.GetMessageType(), "mac", mac, "err", err).Info("retrieved job is empty")
+		metrics.JobsInProgress.With(labels).Dec()
+		timer.ObserveDuration()
 		return
 	}
-	go j.ServeDHCP(w, req)
+	go func() {
+		if j.ServeDHCP(w, req) {
+			metrics.DHCPTotal.WithLabelValues("reply", gi.String()).Inc()
```
mmlb, line 83: why don't we want to explicitly record total vs successful and thus be able to calculate error'd?

metrics/setup.go
```diff
@@ -0,0 +1,142 @@
...
+	labelValues := []prometheus.Labels{
+		{"op": "DHCPACK", "giaddr": "0.0.0.0"},
+		{"op": "DHCPDECLINE", "giaddr": "0.0.0.0"},
+		{"op": "DHCPDISCOVER", "giaddr": "0.0.0.0"},
+		{"op": "DHCPINFORM", "giaddr": "0.0.0.0"},
+		{"op": "DHCPNAK", "giaddr": "0.0.0.0"},
+		{"op": "DHCPOFFER", "giaddr": "0.0.0.0"},
+		{"op": "DHCPRELEASE", "giaddr": "0.0.0.0"},
+		{"op": "DHCPREQUEST", "giaddr": "0.0.0.0"},
+		{"op": "reply", "giaddr": "0.0.0.0"},
```
mmlb, line 41: reply? what does this mean/how is it used?

mmlb: ok looks like this is meant for "successful" replies. But that should probably be handled via the normal dhcp response right? We should never see a DHCPOFFER coming in since we config dhcp4-go to ignore w/e we don't care about anyway.

I'm back to thinking we need to have total and responded/error separately.

metrics/setup.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	initCounterLabels(DHCPTotal, labelValues)
+
+	CacherDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "cacher_request_duration_seconds",
+		Help:    "Duration of cacher requests",
```
mmlb, line 47: missing the `.` at the end

metrics/setup.go
```diff
@@ -0,0 +1,142 @@
...
+		{"from": "ip"},
+	}
+	initObserverLabels(CacherDuration, labelValues)
+	initCounterLabels(CacherCacheHits, labelValues)
+	initCounterLabels(CacherTotal, labelValues)
+	initGaugeLabels(CacherRequestsInProgress, labelValues)
+
+	DiscoverDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "discover_duration_seconds",
+		Help:    "Duration taken to get a responce for a newly discovered request",
```
mmlb, line 74: `.`

dhcp.go
mikemrm: ditto response?

dhcp.go
mikemrm: We're able to do that with reply.

metrics/setup.go
mikemrm: I thought the same, however I thought it would be useful to not update DHCPOFFER in case a DHCPOFFER made it to the server. Seeing that number go up could help determine that there are multiple dhcp servers on the network.

I can add a dhcp_offers generic count separately. But since this one tracks by giadder, it might be good to keep this here.

mikemrm: ah just noticed dhcp_totals is also broken out by giadder, so i could do the same as dhcp_totals, and remove reply from here

dhcp.go
mmlb: derp! makes sesnse!

metrics/setup.go
mmlb: maybe add a direction then?

dhcp.go
mmlb: ditto derpiness

metrics/setup.go
mikemrm: switched to making op = send/recv and moved the type to a type tag.

dhcp.go
mikemrm: switched to using op=send and type=DHCPOFFER

metrics/setup.go
```diff
@@ -0,0 +1,142 @@
...
+	labelValues := []prometheus.Labels{
+		{"op": "recv", "type": "DHCPACK", "giaddr": "0.0.0.0"},
+		{"op": "recv", "type": "DHCPDECLINE", "giaddr": "0.0.0.0"},
+		{"op": "recv", "type": "DHCPDISCOVER", "giaddr": "0.0.0.0"},
+		{"op": "recv", "type": "DHCPINFORM", "giaddr": "0.0.0.0"},
+		{"op": "recv", "type": "DHCPNAK", "giaddr": "0.0.0.0"},
+		{"op": "recv", "type": "DHCPOFFER", "giaddr": "0.0.0.0"},
+		{"op": "recv", "type": "DHCPRELEASE", "giaddr": "0.0.0.0"},
+		{"op": "recv", "type": "DHCPREQUEST", "giaddr": "0.0.0.0"},
+		{"op": "send", "type": "DHCPOFFER", "giaddr": "0.0.0.0"},
```
mmlb, line 41: missing `{"op": "send", "type": "DHCPACK", "giaddr": "0.0.0.0"},`
also :thinking: how does a different giaddr not blow up?

mikemrm: This list will definitely grow. But it should typically only grow to as many relays as you have pointed to it. I thought this would be a good way to potentially see hotspots in requests.

PR #26, opened 2020-04-30
Add VMware ESXi 7.0

dlaube:
The purpose of this pull request is to add initial supportfor ESXi 7.0 in the boots vmware installer.


ScottGarman:
Changes look sane based on my limited knowledge thus far. I think the 3 commits should get squashed into one though.

dlaube:
> Changes look sane based on my limited knowledge thus far. I think the 3 commits should get squashed into one though.

Will squash on merge for this.

PR #27, opened 2020-05-08
Gauravgahlot/new data model

kdeng3849:
Ref: [RFC #2 New Hardware data model](https://github.com/tinkerbell/tink/wiki/RFC-%232-New-Hardware-data-model)
 
Abstracting type Discovery into an interface to settle the differences between the new and old hardware data models.

Changes:
- Added an envvar `DATA_MODEL_VERSION` to switch between the old cacher model and the new tinkerbell model
  - if envvar is unset/empty/anything other than `1`, it will default to using the old cacher data model
- Abstracted `Discovery` and `Hardware` among others to take into account new data model

TODO:
- [ ] remove the `replace` in `go.mod` pointing to my fork
- [x] squash some of the commits

Related PRs:
hegel: https://github.com/tinkerbell/hegel/pull/13
tink: https://github.com/tinkerbell/tink/pull/64


job/job.go
```diff
@@ -100,55 +100,39 @@ func (j Job) MarkDeviceActive() {
 //	j.postFailure(reason)
 //}
 
-func (j *Job) setup(d *packet.Discovery) error {
-	mode, netConfig := d.Mode(j.mac), d.NetConfig(j.mac)
-	j.Logger = joblog.With("mac", j.mac, "hardware.id", d.Hardware.ID)
+func (j *Job) setup(dp *packet.Discovery) error {
```
yetang-equinix, line 103: It will be nice to have a comments for this function, what will it do? processing dhcp discovery packet? not looks like it .

packet/client.go
```diff
@@ -10,25 +9,22 @@ import (
...
-	authToken     string
-	cacher        hardwareGetter
+	http           *http.Client
+	baseURL        *url.URL
+	consumerToken  string
+	authToken      string
+	hardwareClient hardwareGetter
 }
 
 func NewClient(consumerToken, authToken string, baseURL *url.URL) (*Client, error) {
```
yetang-equinix, line 30: Is this for DHCP client? wondering why we need client functions here in boots

kdeng3849: The `hardwareClient` refers to the tink/cacher clients responsible for getting the hardware data, see the usage here: https://github.com/tinkerbell/boots/blob/8ecd0d6d08f11d4d6e4bdb9c6b4893815e060ac9/packet/client.go#L46

yetang-equinix: as discussed move to workflow

packet/models_tinkerbell.go
```diff
@@ -0,0 +1,160 @@
+package packet
```
yetang-equinix, line 1: We have models.go, models_cacher.go, models_tinkerbell.go. As we discussed, it will nice to have file header comments to briefly describe which is for what.

PR #31, opened 2020-05-19
Add ECLYPSIUM_TOKEN environment variable | ENG-5336

splaspood:
To enable eclypsium scanning/registration during deprovision but not expose the
token in git we allow the token to be set during run time via an environment
variable and then pass the token into osie during boot.

* add ECLYPSIUM_TOKEN environment variable
* supply eclypsium_token kernel param during osie if defined

### DIFF
--- a/conf/config.go
+++ b/conf/config.go
@@ -27,6 +27,9 @@ var (
 	ignoredGIs  = getIgnoredGIs()
 
 	TrustedProxies = parseTrustedProxies()
+
+	// Eclypsium registration token, passed into osie
+	EclypsiumToken = env.Get("ECLYPSIUM_TOKEN")
 )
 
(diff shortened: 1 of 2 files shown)


conf/config.go
```diff
@@ -27,6 +27,9 @@ var (
 	ignoredGIs  = getIgnoredGIs()
 
 	TrustedProxies = parseTrustedProxies()
+
+	// Eclypsium registration token, passed into osie
+	EclypsiumToken = env.Get("ECLYPSIUM_TOKEN", "")
```
mmlb, line 32: just `env.Get("ECLYPSIUM_TOKEN")` is best.

splaspood: ack.

mmlb:
lgtm, please squash/rebase.

PR #40, opened 2020-06-12
Custom osie via userdata

mmlb:
This lets us specify a custom osie version (for testing, yay testing) by supplying the info at provision time, via userdata. I've set it up so that when API has support for this in the provision call it should work already.


ScottGarman:
I'm not well-versed enough in the boots codebase to give this a thorough review, but I like the looks of what I see and am :100: to having this available soonish. :+1:

PR #44, opened 2020-06-23
Cleanup after new data model

mmlb:
## Description

Change from pointers to interface to just an interface for the recent data model split. Took the opportunity to clean up some of the related tests so they would be more correct (when setting env vars) / easier to follow (making TestNewDiscoveryMismatch not panic). 

## Why is this needed

I think #27 was merged prematurely. Lots of commented out code that should have been deleted, oddness in the pointer-to-interface and some hard to follow tests. This is me polishing some of that up.

Fixes: #

## How Has This Been Tested?

`go test ./...`

No anticipated impact.

## Checklist:

I have:

- [x] added unit or e2e tests


mmlb:
As always, reviewing each commit one-at-a-time would make it easiest to review.

nathangoulding:
I've gone through every commit, and they all look good to me. Thanks for this. 👍

PR #46, opened 2020-06-23
Set different firstdisk on c3.medium.x86.01 plan version

dlaube:
This is to add logic around the firstdisk config option, specifically for the v1 plan version of c3.medium.


mmlb:
really wish we had some sort of metadata for the images where we can grab this from

PR #47, opened 2020-06-24
Fixes #45: dhcp lease time issues

### LINKED ISSUE #45: dhcp lease time issues
With the introduction of new data model, it does not use the environment variable for default DHCP lease time. In addition, there is not a default lease time. This causes the value of 0 to be used, which creates a flood of DHCP requests to the DHCP server because of the short lease time. We should default this to 86400 if the user does not supply one.

nathangoulding:
Resolves #45


packet/models_tinkerbell.go
```diff
@@ -14,19 +16,29 @@ func (i InterfaceTinkerbell) Name() string {
 }
 
 func (d DiscoveryTinkerbellV1) LeaseTime(mac net.HardwareAddr) time.Duration {
-	return d.Network.InterfaceByMac(mac).DHCP.LeaseTime
+	leaseTime := d.Network.InterfaceByMac(mac).DHCP.LeaseTime
+	fmt.Println(leaseTime)
```
rgl, line 20: You forgot to remove this `fmt.Println` (and the others) logging lines.

nathangoulding: PTAL

Issue #35, opened 2020-05-28
boots should not PXE if no workflow defined


nathangoulding:
If there is no active workflow for a worker, boots should not PXE boot a system. Right now, it will always PXE into the environment and noop, which is not a practical use case.

rgl:
Instead of being a noop I would like to be able to specify a workflow for the machines that do not (yet) have a workflow. That way I can deal with them (e.g. run some kind of inventory application on them and show the machine in some kind of "un-provisioned/new" machines catalog).

nathangoulding:
@rgl Agreed this would be a good capability to create a default "wait-for-work" template + container that comes baked with Tinkerbell. We also need this in Packet, and I think would have broad appeal. Perhaps we can capture that in a separate issue.

Cbkhare:
Updates so far. 

-) Development & Testing activity:

1) Changes to fetch worker id for given hardware. Dev + test  (Done)

2) Changes to fetch workflow for a given worker id. Dev - (Done) + Test in progress. 

3) Point 2 requires addition of a rpc client to fetch workflow, changes is required in both boots and tink. Actually, I was using existing client to fetch workflow and after i got go panic error i found, rpc client is currently specific to hardware clis. So, i added a rpc client for workflow. Dev  (Done) + Boots Test  (Done) + Tink in progress. 

4) Endpoints to fetch workflow using client. Dev (Done) + Test, in progress, currently fixing client in point3.

5) Using above workflow client, i can fetch workflows and filter which are active. Once, my testing on clients and endpoints are completed i will complete this. Dev + Test (in progress) 


I have raised below PR, for workflow client changes in boots and tink.  Later, i will add my other changes in same branch. 

https://github.com/tinkerbell/boots/pull/50/commits/96bb35d1c7a21e8d6e9c18e4dac98aecbc83ab7c
https://github.com/tinkerbell/tink/pull/200/commits/f14f0700511b8482a349e3dcc99a23132f31f82c

Cbkhare:
Tested this PR along with https://github.com/tinkerbell/tink/pull/200. 

- Testing with Active workflow :- completed 

[...]

PR #50, opened 2020-07-01
No pxe boot #35

Cbkhare:
## Details 
No pxe boot is to be supported if there are no active workflow for a worker/hardware. 

## Changes
Added endpoint to fetch active workflow for a hardware
Added logic to translate hardware id to active worker id
Added logic to stop pxe boot if not active workflow is present. 

## Testing
Tested with client added via https://github.com/tinkerbell/tink/pull/200. 

happy path testing completed. 

`{"level":"info","ts":1593695857.2480636,"caller":"job/job.go:55","msg":"found active workflow for hardware 0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94 with workflow id d0f730bc-894d-42de-8c37-9e423f5c761d","service":"github.com/tinkerbell/boots","pkg":"http","dhcp":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94"}`

Need to do testing with new tink after the merge of above mentioned PR. 

Negative testing pending.


packet/client.go
```diff
@@ -62,6 +73,7 @@ func NewClient(consumerToken, authToken string, baseURL *url.URL) (*Client, erro
 		if err != nil {
 			return nil, errors.Wrap(err, "connect to cacher")
 		}
+		wg = hg
```
kdeng3849, line 76: I don't think it makes sense to set the cacher client to both the `client.hardwareClient` and the `client.workflowClient`, since cacher wouldn't be dealing with workflows. And I'm also pretty sure it would panic if you tried to cast a cacher client into a tink `WorkflowSvcClient`.  
I think it's better to just leave it as nil and check for nil before fetching for workflows.

Cbkhare: done, pointed it to nil.

job/job_test.go
```diff
@@ -173,3 +175,46 @@ func TestSetupFails(t *testing.T) {
 	// should still be able to log, see #_incident-130
 	j.With("happyThoughts", true).Error(err)
 }
+
+func TestDiscoverActiveWorkflowForMacFound(t *testing.T) {
```
gianarb, line 179: Can you write an assertion that checks the actual return? For example, can you assert that `status` is `true`?

job/job.go
```diff
@@ -34,6 +35,34 @@ type Job struct {
...
+		return false, err
+	}
+
+	// wcl is workflowContextList, for more details refer to workflow.pb.go in tink
+	for _, wf := range wcl.WorkflowContexts {
+		if wf.CurrentActionState == 0 {
+			active = true
+			s := fmt.Sprintf("found active workflow for hardware %s with workflow "+
+				"id %s", hwID, wf.WorkflowId)
+			joblog.With("dhcp", hwID).Info(s)
```
gianarb, line 59: If you `return true` here I think you do not need to break the loop. And you can `return false` at the end. Removing the allocation at line 43.

I think I am also not 100% sure about the name/doc block for this function.

is this function returning `true` if the given Hardware has an active workflow? If so can we call this function something like: `isHardwareExecutingWorkflow`? I tend to use `is` as a prefix when a function returns `true/false` because it is a bit more sematic.

job/job_test.go
```diff
@@ -173,3 +175,46 @@ func TestSetupFails(t *testing.T) {
...
+	}
+	status, err := discoverActiveWorkflowForMac(hwID)
+	if err != nil {
+		t.Fatal("error occured while testing")
+	}
+	t.Logf("status of disovery action is %t", status)
+}
+
+func TestDiscoverActiveWorkflowForMacNotFound(t *testing.T) {
+	var hwID string = "0eba0bf8-fake-4b4a-ab9f-6ebe93b90a94"
```
gianarb, line 201: Same here, can you assert that the returned value is `status==false`?

job/job.go
```diff
@@ -34,6 +35,34 @@ type Job struct {
...
+
+	s := fmt.Sprintf("fetching workflows for hardware %s", hwID)
+	joblog.With("dhcp", hwID).Info(s)
+
+	wcl, err := tinkWorkflowClient(hwID)
+	if err != nil {
+		return false, err
+	}
+
+	// wcl is workflowContextList, for more details refer to workflow.pb.go in tink
```
mmlb, line 53: this comment seems unnecessary, any editor should be able to provide this same info

job/job.go
mmlb, line 59: these args should really be kv pairs in the log line like:

```go
l.With("worflowID", wf.WorkflowId).Info("found active workflow for hardware")
```

job/job.go
```diff
@@ -34,6 +35,34 @@ type Job struct {
...
 
+var tinkWorkflowClient = client.GetWorkflowsFromTink
+
+// discoverWorkflowForMac fetches the workflow for a given MAC
+// hwID is the hardware/worker ID corresponding to the MAC
+func discoverActiveWorkflowForMac(hwID string) (bool, error) {
+	var active bool = false
+
+	s := fmt.Sprintf("fetching workflows for hardware %s", hwID)
+	joblog.With("dhcp", hwID).Info(s)
```
mmlb, line 46: this should be:
```go
l := joblog.With("hardwareID", hwID)
l.Info("fetching workflows for hardware")
```

job/job.go
mmlb: Agree with all of this

job/job.go
```diff
@@ -50,6 +79,20 @@ func CreateFromDHCP(mac net.HardwareAddr, giaddr net.IP, circuitID string) (Job,
 	if err != nil {
 		return Job{}, err
 	}
+
+	// check if there is an active workflow for the device
+	hd := (*d).Hardware()
+	hwID := (*hd).HardwareID()
+	activeWorkflows, e := discoverActiveWorkflowForMac(hwID)
+	if e != nil {
+		return Job{}, e
```
mmlb, line 88: please use `err`, not `e`

job/job.go
```diff
@@ -50,6 +79,20 @@ func CreateFromDHCP(mac net.HardwareAddr, giaddr net.IP, circuitID string) (Job,
...
+	}
+
+	if !activeWorkflows {
+		s := fmt.Sprintf("No active workflow found for hardware %s", hwID)
```
mmlb, line 92: `No` should be `no`

job/job.go
```diff
@@ -84,6 +127,20 @@ func CreateFromIP(ip net.IP) (Job, error) {
...
 		return Job{}, err
 	}
+
+	// check if there is an active workflow for the device
+	hd := (*d).Hardware()
+	hwID := (*hd).HardwareID()
+	activeWorkflows, e := discoverActiveWorkflowForMac(hwID)
+	if e != nil {
+		return Job{}, e
+	}
```
mmlb, line 137: ditto

job/job.go
```diff
@@ -84,6 +127,20 @@ func CreateFromIP(ip net.IP) (Job, error) {
...
+
+	if !activeWorkflows {
+		s := fmt.Sprintf("No active workflow found for hardware %s", hwID)
```
mmlb, line 140: ditto

job/job.go
```diff
@@ -34,6 +35,34 @@ type Job struct {
 	instance *packet.Instance
 }
 
+var tinkWorkflowClient = client.GetWorkflowsFromTink
```
mmlb, line 38: please no global variable, just pass in a client to `discoverActiveWorkflowForMac`

job/job.go
```diff
@@ -34,6 +35,34 @@ type Job struct {
...
+
+// discoverWorkflowForMac fetches the workflow for a given MAC
+// hwID is the hardware/worker ID corresponding to the MAC
+func discoverActiveWorkflowForMac(hwID string) (bool, error) {
```
mmlb, line 42: please use `MAC` instead of `Mac` in the name and comment.

mmlb: Actually nm, lets go with @gianarb's comment since it makes more sense and this func doesn't even use MAC itself anyway.

packet/client.go
```diff
@@ -62,6 +73,7 @@ func NewClient(consumerToken, authToken string, baseURL *url.URL) (*Client, erro
 		if err != nil {
 			return nil, errors.Wrap(err, "connect to cacher")
 		}
+		wg = nil
```
mmlb, line 76: this assignment is ineffective, wg is already nil here anyway.

packet/endpoints.go
```diff
@@ -32,6 +33,47 @@ type ComponentsResponse struct {
...
+		return result, errors.New("missing hardware id")
+	}
+
+	labels := prometheus.Labels{"from": "dhcp"}
+	cacherTimer := prometheus.NewTimer(metrics.CacherDuration.With(labels))
+	metrics.CacherRequestsInProgress.With(labels).Inc()
+	metrics.CacherTotal.With(labels).Inc()
+
+	if c.workflowClient == nil {
+		return result, errors.New("Unable to fetch workflow, client does not exist")
```
mmlb, line 51: Unable should be lower case `u`

packet/endpoints.go
```diff
@@ -32,6 +33,47 @@ type ComponentsResponse struct {
...
+	}
+
+	j, err := json.Marshal(resp)
+	if err != nil {
+		return result, errors.New("marshalling tink hardware")
+	}
+
+	err = json.Unmarshal(j, &result)
+	if err != nil {
+		return result, errors.New("Unmarshalling tink workflows")
```
mmlb, line 71: `U` -> `u`

job/job.go
Cbkhare: My main objective of having this variable is to mock client.GetWorkflowsFromTink in unit test under job/job_test.go. If we want to remove this, Do you think that can be done in other way ?

job/job.go
Cbkhare: i have renamed this method.

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done

packet/endpoints.go
Cbkhare: done

job/job_test.go
Cbkhare: done

job/job_test.go
Cbkhare: done

job/job.go
Cbkhare: done, renamed it to `isHardwareHavingActiveWorkflow`

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done

packet/client.go
Cbkhare: done, removed

packet/endpoints.go
Cbkhare: done

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done

Cbkhare:
> I tend to write a single `Test` that use Table Testing https://github.com/golang/go/wiki/TableDrivenTests when I can because it avoids the retention of the logic that we are testing. For example in this case we can have a map of 3 tests:
> 
> ```
> 1. Hardware not found because it does not exists
> 2. no active workflow for an existing hardwareID should return false
> 3. active workflow for an existing hardwareID should return true
> ```
> 
> Let me know what you think about this

@gianarb , This strategy does looks good, but IMO i think at first we should refactor our code to support mocking of methods which are called inside our TestMethod. I am not sure, how TableDrivenTests will mock methods called inside test method. Also we may have to consider cases, as in this PR, where input is same but test result differs on the basis of output obtained from mocked method.

job/job.go
mmlb: Yep change `discoverActiveWorkflowForMac(hwID)` to be `discoverActiveWorkflowForMac(client, hwID)` and then all the callers need to pass in what they want. Normal code can do `client.GetWorkflowsFromTink`, test code can do the mock.

Cbkhare:
Test results and boot logs after changes with review comments 

with active workflow
`{"level":"info","ts":1594301567.8052936,"caller":"job/job.go:62","msg":"fetching workflows for hardware 0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","service":"github.com/tinkerbell/boots","pkg":"http","hardwareID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94"}
{"level":"info","ts":1594301567.81777,"caller":"job/job.go:71","msg":"found active workflow for hardware","service":"github.com/tinkerbell/boots","pkg":"http","hardwareID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","workflowID":"6d360928-2b7b-4ec3-b5ea-530de2c8e7cf"}`

No active workflow 
`"fetching workflows for hardware 0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","service":"github.com/tinkerbell/boots","pkg":"http","hardwareID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94"}
{"level":"info","ts":1594301723.3819473,"caller":"boots/dhcp.go:76","msg":"retrieved job is empty","service":"github.com/tinkerbell/boots","pkg":"main","type":"DHCPDISCOVER","mac":"08:00:27:00:00:01","err":"no active workflow found for hardware 0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94"`

packet/endpoints.go
Cbkhare: i found it  while addressing review comments, have fixed it

packet/client.go
Cbkhare: done

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done

job/job.go
Cbkhare: @mmlb i have added interfaces to mock `client.GetWorkflowsFromTink`. Can you please confirm if the changes are right

job/job.go
```diff
@@ -50,6 +91,20 @@ func CreateFromDHCP(mac net.HardwareAddr, giaddr net.IP, circuitID string) (Job,
...
+	hd := (*d).Hardware()
+	hwID := (*hd).HardwareID()
+	activeWorkflows, err := isHardwareHavingActiveWorkflow(hwID)
+	if err != nil {
+		return Job{}, err
+	}
+
+	if !activeWorkflows {
+		s := fmt.Sprintf("no active workflow found for hardware %s", hwID)
+		return Job{}, errors.New(s)
```
mmlb, line 105: can be combined into just `errors.Errorf`

job/job.go
```diff
@@ -84,9 +139,24 @@ func CreateFromIP(ip net.IP) (Job, error) {
...
```
mmlb, line 153: can be combined into just `errors.Errorf`

job/job.go
```diff
@@ -9,9 +10,17 @@ import (
...
 
-var client *packet.Client
+var (
+	client    *packet.Client
+	connectTo endpointClient
+)
+
+func init() {
+	connectTo = &connect{}
+}
```
mmlb, line 23: This is still a global var. And also now its a new "large" interface, when the just a func was better imo. Why not just make `isHardwareActiveWorkflow` take in a function?

job/job.go
```diff
@@ -34,6 +43,38 @@ type Job struct {
...
+
+	wcl, err := connectTo.getWorkflowsFromTink(hwID)
+	if err != nil {
+		return false, err
+	}
+
+	for _, wf := range wcl.WorkflowContexts {
+		if wf.CurrentActionState == 0 {
+			joblog.With("hardwareID", hwID, "workflowID", wf.WorkflowId).Info("found" +
+				" active workflow for hardware")
```
mmlb, line 71: just on one line please

packet/endpoints.go
```diff
@@ -32,6 +33,47 @@ type ComponentsResponse struct {
...
+	}
+
+	j, err := json.Marshal(resp)
+	if err != nil {
+		return result, errors.New("marshalling tink hardware")
+	}
+
+	err = json.Unmarshal(j, &result)
+	if err != nil {
+		return result, errors.New("unmarshalling tink workflows")
```
mmlb, line 71: you should not be returning result here since we don't know if `json.Unmarshal` may have set any fields and we should only return empty or valid objects.

packet/endpoints.go
```diff
@@ -32,6 +33,47 @@ type ComponentsResponse struct {
...
+	if err != nil {
+		return result, errors.New("fetching the workflow")
+	}
+
+	j, err := json.Marshal(resp)
+	if err != nil {
+		return result, errors.New("marshalling tink hardware")
+	}
+
+	err = json.Unmarshal(j, &result)
```
mmlb, line 69: This pattern seems very weird to me, why are using json as an intermediate representation?

Cbkhare: I was observing error on return type, let me retry this.

packet/endpoints.go
Cbkhare: same as below

job/job.go
Cbkhare: shouldn't we  be following 80 character line width rule ?

job/job.go
Cbkhare: I have modified the method to pass `connectTo` in method. 

Earlier, it was single method so passing the method as parameter could have been fine. But as the complexity increases, number of methods to be mocked also increases, interface would be eventually a better option for testing.

packet/endpoints.go
Cbkhare: done modified.

job/job.go
Cbkhare: done

Cbkhare:
Will do cherry pick once review is completed.

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done

packet/client.go
```diff
@@ -19,12 +19,16 @@ import (
 type hardwareGetter interface {
 }
 
+type workflowGetter interface {
+}
```
mmlb, line 23: this should not be an empty interface

job/job_test.go
```diff
@@ -173,3 +176,55 @@ func TestSetupFails(t *testing.T) {
 	// should still be able to log, see #_incident-130
 	j.With("happyThoughts", true).Error(err)
 }
+
+type mockJob struct{}
```
mmlb, line 180: All this new test code can be split off into a separate test func and then we can easily make it a table-driven test by having the `tw.WorkflowContextList` be part of the table. Another reason to make `isHardwareHavingActiveWorflow` take in the `tw.WorkflowContextList` instead of a fetcher.

packet/endpoints.go
```diff
@@ -32,6 +33,50 @@ type ComponentsResponse struct {
 	Components []Component `json:"components"`
 }
 
+// GetWorkflowsFromTink fetches the list of workflows from tink
+func (c *Client) GetWorkflowsFromTink(hwID string) (*tw.WorkflowContextList, error) {
+
+	var result *tw.WorkflowContextList
```
mmlb, line 39: this should be defined closer to where its *actually* used and returned, not up here. All the error cases should `return nil, err...`

job/job.go
```diff
@@ -34,6 +38,38 @@ type Job struct {
...
+}
+
+func (co *connect) getWorkflowsFromTink(hwID string) (*tw.WorkflowContextList, error) {
+	return client.GetWorkflowsFromTink(hwID)
+}
+
+// isHardwareHavingActiveWorkflow fetches workflows for a given hwID and returns
+// the status true if there is a pending (active) workflow for hwID
+// hwID is the hardware/worker ID corresponding to the MAC
+func isHardwareHavingActiveWorkflow(connectTo endpointClient, hwID string) (bool, error) {
```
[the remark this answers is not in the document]
Cbkhare: Having separate method to fetch the result would make redundant method to do the same thing. I would like to emphasise interface can not be removed. Again even if we separate the responsibility, to unit test the fetcher you will need interface.

packet/endpoints.go
Cbkhare: i will remove this, but ideally all variable in a method should be defined in a single place and preferably onto the top.

job/job_test.go
Cbkhare: :) in that case we will have to add separate unit test for fetcher which will use existing structs of type mocks to complete the testing. I mean in that, `isHardwareHavingActiveWorflow` will lose it sole purpose and having it unit tested will be minuscule compared to fetcher method.

packet/client.go
Cbkhare: any suggestion, what we can place in here ?

packet/endpoints.go
```diff
@@ -32,6 +33,50 @@ type ComponentsResponse struct {
...
+				return result, errors.New("marshalling tink hardware")
+			}
+
+
+		err = json.Unmarshal(resp, &result)
+		if err != nil {
+			return result, errors.New("unmarshalling tink workflows")
+		}
+	*/
+
```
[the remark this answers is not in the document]
Cbkhare: done

job/job.go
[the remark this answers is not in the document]
Cbkhare: i have segregated the call to fetch workflow for a hardware.

job/job_test.go
Cbkhare: added tabledriven test

packet/endpoints.go
Cbkhare: have placed the variables in the method signature

packet/client.go
gianarb: I understand why you need it but it is not a good practice. An empty interface means EVERYTHING. And that's too much. If we do not have a mock for the `workflowClient` we should probably create one. Or either way, the `workflowGetter` and the `hardwareGetter` should require the functions used by the underline components.

Cbkhare: @gianarb yes, make sense. I have added get set methods in interface. Currently this is only for workflow client. We can also use this to mock the client. If this works fine maybe we can do this for hardware client as well. 

commit : https://github.com/tinkerbell/boots/pull/50/commits/375a7166ea03e2a0266a822ac6078f1a9883fe80

Cbkhare:
have to sqash commit. waiting for review to complete.

packet/client.go
gianarb: Yep, in my opinion, this works way better! Thanks

job/job.go
```diff
@@ -34,6 +37,20 @@ type Job struct {
 	instance *packet.Instance
 }
 
+// isHardwareHavingActiveWorkflow fetches workflows for a given hwID and returns
+// the status true if there is a pending (active) workflow for hwID
+// hwID is the hardware/worker ID corresponding to the MAC
+func isHardwareHavingActiveWorkflow(wcl *tw.WorkflowContextList) (bool, error) {
+
+	for _, wf := range (*wcl).WorkflowContexts {
+		if wf.CurrentActionState == 0 {
```
mmlb, line 46: Please don't use a magic number here, use the ActionState enum

job/job.go
```diff
@@ -34,6 +37,20 @@ type Job struct {
 	instance *packet.Instance
 }
 
+// isHardwareHavingActiveWorkflow fetches workflows for a given hwID and returns
+// the status true if there is a pending (active) workflow for hwID
+// hwID is the hardware/worker ID corresponding to the MAC
+func isHardwareHavingActiveWorkflow(wcl *tw.WorkflowContextList) (bool, error) {
+
```
mmlb, line 44: please remove this extra line.

job/job.go
```diff
@@ -84,9 +121,30 @@ func CreateFromIP(ip net.IP) (Job, error) {
 	if err != nil {
 		return Job{}, err
 	}
+
+	// check if there is an active workflow for the device
```
mmlb, line 125: this comment doesn't really make sense on this line

job/job_test.go
```diff
@@ -173,3 +175,41 @@ func TestSetupFails(t *testing.T) {
 	// should still be able to log, see #_incident-130
 	j.With("happyThoughts", true).Error(err)
 }
+
+func TestIsHardwareHavingActiveWorkflow(t *testing.T) {
+
```
mmlb, line 180: please remove empty line

packet/client.go
```diff
@@ -14,17 +14,42 @@ import (
...
+	if err != nil {
+		return err
+	}
+	(*w).wsc = wsc
+	return nil
+}
+
+func (w *workflow) getWorkflowConnection() tw.WorkflowSvcClient {
+	return w.wsc
+}
```
mmlb, line 43: we no longer need any of this right? Now that we are just passing the wcl into `isHardwareHavingActiveWorkflow` and we don't have different concrete types. Better to get rid of all this unneeded code then.

job/job.go
```diff
@@ -34,6 +37,20 @@ type Job struct {
 	instance *packet.Instance
 }
 
+// isHardwareHavingActiveWorkflow fetches workflows for a given hwID and returns
+// the status true if there is a pending (active) workflow for hwID
+// hwID is the hardware/worker ID corresponding to the MAC
+func isHardwareHavingActiveWorkflow(wcl *tw.WorkflowContextList) (bool, error) {
```
mmlb, line 43: let rename this to `hasActiveWorkflow` its shorter and easier to grok.

packet/client.go
Cbkhare: It is not related `isHardwareHavingActiveWorkflow`. You may want to check recent updates in endpoints.go. 

This method passes the tink workflow client to `GetWorkflowsFromTink` in endpoints.go.

job/job.go
Cbkhare: 0 is harcoded for pending on workflow.pb.go, do you want to replace the '0' with constant from workflow.pb.go , `ActionState_ACTION_PENDING`  or use map for `ACTION_PENDING` `ActionState_value` ?

job/job_test.go
Cbkhare: done

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done

job/job.go
Cbkhare: done.

packet/client.go
Cbkhare: As discussed, I have kept `getWorkflowConeection` and have removed `setWorkflowConnection`

packet/client.go
```diff
@@ -14,17 +14,32 @@ import (
...
+	getWorkflowConnection() tw.WorkflowSvcClient
+}
+
+type workflow struct {
+	wsc tw.WorkflowSvcClient
+}
+
+func (w *workflow) getWorkflowConnection() tw.WorkflowSvcClient {
+	return w.wsc
+}
```
mmlb, line 33: Hey @Cbkhare I'm back to not wanting this interface here. When you and I talked offline you convinced me it would be useful for testing the `GetWorkflowsFromTink` and `DiscoverHardwareFrom*` methods. But it still was bugging me since the grpc client stuff is already an interface and it seemed like we could use a mock implementation of it to do the testing, but I was going to drop it.

But then I went and wanted to throw in some tests in #59 and went looking for mocking grpc clients in go and found that we can just mock the CacherClient (in my use case) for the tests. Lets go that route instead, PTAL at #59 for the nitty gritty.

Cbkhare: done, as discussed offline today, have removed the interface and added the call's as below 

in `packet/client.go`
```
import tw "github.com/tinkerbell/tink/protos/workflow"
# client struct {
..
workflowClient tw.WorkflowSvcClient //  instead of workflowGetter
}
// inside func NewClient
wg, err = tinkClient.TinkWorkflowClient()
if err != nil {
	return nil, errors.Wrap(err, "connect to tink")
}
```
in `packet/endpoints.go`
```
// c is the *client 
result, err = c.workflowClient.GetWorkflowContexts(context.Background(), &tw.WorkflowContextRequest{WorkerId: hwID}) 
```

job/job.go
```diff
@@ -9,9 +9,12 @@ import (
 	"github.com/tinkerbell/boots/conf"
 	"github.com/tinkerbell/boots/dhcp"
 	"github.com/tinkerbell/boots/packet"
+	tw "github.com/tinkerbell/tink/protos/workflow"
 )
 
-var client *packet.Client
+var (
+	client *packet.Client
+)
```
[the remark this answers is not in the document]
Cbkhare: done

packet/endpoints.go
```diff
@@ -32,6 +33,33 @@ type ComponentsResponse struct {
 	Components []Component `json:"components"`
 }
 
+// GetWorkflowsFromTink fetches the list of workflows from tink
+func (c *Client) GetWorkflowsFromTink(hwID string) (result *tw.WorkflowContextList, err error) {
```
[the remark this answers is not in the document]
Cbkhare: done

packet/endpoints.go
```diff
@@ -32,6 +33,33 @@ type ComponentsResponse struct {
...
+
+	labels := prometheus.Labels{"from": "dhcp"}
+	cacherTimer := prometheus.NewTimer(metrics.CacherDuration.With(labels))
+	metrics.CacherRequestsInProgress.With(labels).Inc()
+	metrics.CacherTotal.With(labels).Inc()
+
+	if c.workflowClient == nil {
+		return result, errors.New("unable to fetch workflow, client does not exist")
+	}
+
```
mmlb, line 50: this is no longer needed right?

packet/endpoints_test.go
```diff
@@ -92,3 +94,54 @@ func TestDiscoverHardwareFromDHCP(t *testing.T) {
...
+			cMock := workflowMock.NewMockWorkflowSvcClient(ctrl)
+			cMock.EXPECT().GetWorkflowContexts(gomock.Any(), gomock.Any()).Return(hw.wcl, hw.err)
+			c.workflowClient = cMock
+		}
+		w, err := c.GetWorkflowsFromTink(hw.hwID)
+		if hw.err != nil {
+			assert.Error(t, err)
+			continue
+		}
+		assert.Equal(t, w, hw.wcl)
```
mmlb, line 145: please wrap in a `t.Run(...)` subtest call. This way if a test fails we get better info, and can even run just that test when doing `-run`

job/job_test.go
```diff
@@ -173,3 +175,40 @@ func TestSetupFails(t *testing.T) {
...
+		},
+	}
+	for _, list := range WorkflowContextLists {
+		s, err := hasActiveWorkflow(list.wcl)
+		if err != nil {
+			t.Fatal("error occured while testing")
+		}
+		if s != list.status {
+			t.Fatalf("failed! expected %t found %t", list.status, s)
+		}
```
mmlb, line 212: lets wrap this in a t.Run too

mmlb:
@Cbkhare aaaaaaaaaalmost there! Thanks for working through all the progress!

mmlb:
Scouts honor, that would be the last changes :D

packet/endpoints_test.go
Cbkhare: done,

packet/endpoints.go
Cbkhare: not sure, but removed it.

job/job_test.go
Cbkhare: done

Cbkhare:
@mmlb @gianarb 
I have addressed all the review comments.

mmlb:
@kdeng3849 @gianarb we're now waiting on your final acks.

PR #57, opened 2020-07-09
Production into master

mmlb:
## Description

Bring in changes that were merged into the `production` branch, and thus deployed into Packet's actual production system, back into master. This was done while master was in a state that would break Packet provisioning.

## Why is this needed

With this merged we can get back to having Packet working off of `master` branch.

## How Has This Been Tested?

`go test` passes.
Tested in both YYZ1 and NRT1, provisions/deprovisions working well.

No anticipated issues.

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/packet/models.go
+++ b/packet/models.go
@@ -68,7 +68,7 @@ type Hardware interface {
 	HardwareState() HardwareState
 	HardwareServicesVersion() string
 	HardwareUEFI(mac net.HardwareAddr) bool
-	OsieBaseURL(mac net.HardwareAddr) string
+	OSIEBaseURL(mac net.HardwareAddr) string
 	KernelPath(mac net.HardwareAddr) string
 	InitrdPath(mac net.HardwareAddr) string
 }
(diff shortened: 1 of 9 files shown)


mmlb:
I've tested this in yyz1 and nrt1, looking good.

packet/models.go
```diff
@@ -139,11 +139,12 @@ type Instance struct {
...
-	AlwaysPXE     bool            `json:"always_pxe,omitempty"`
-	IPXEScriptURL string          `json:"ipxe_script_url,omitempty"`
-	IPs           []IP            `json:"ip_addresses"`
-	UserData      string          `json:"userdata,omitempty"`
+	OS              OperatingSystem `json:"operating_system_version"`
+	AlwaysPXE       bool            `json:"always_pxe,omitempty"`
+	IPXEScriptURL   string          `json:"ipxe_script_url,omitempty"`
+	IPs             []IP            `json:"ip_addresses"`
+	UserData        string          `json:"userdata,omitempty"`
+	servicesVersion ServicesVersion
```
kdeng3849, line 152: Is the "s" meant to be lowercase here?

kdeng3849: oh wait I see it

PR #59, opened 2020-07-15
Fix api discovery fetch

mmlb:
## Description

This fixes the DiscoverFromDHCP fallback code that reaches to API if the data is not in cacher. The fallback only makes sense in Cacher mode, so the code returns in the tinkerbell switch branch if no data comes back from tink. In Cacher mode, the Post to Packet API is done with a concrete DiscvoeryCacher type instead of the Discovery interface. Passing the interface was garbage since there's no actual type for json to populate fields off of.

## Why is this needed

Fixes: ENG-7294

## How Has This Been Tested?

YOLO, tested in production.

This is only impacting Packet internally, specifically servers that are not in the Public Provisioning Pool yet.

## Checklist:

I have:

- [x] added unit or e2e tests


mmlb:
@kdeng3849 @jgavinray no longer Draft as it now has tests! :tada: !

PR #69, opened 2020-08-25
use packethost/tftp-go for fixed block size

mikemrm:
## Description

In some cases we hit an issue with mtu and tftp in cilium. This change
includes a fix to reduce the block size to 1400 which has been working
well under cilium.


mikemrm:
switching to using tinkerbell org

PR #71, opened 2020-08-25
check for new datamodel before getting workflows

mikemrm:
Without this check boots is incompatibles with the old version of
cacher. This check skips getting workflows when not using the new
datamodel.


job/job.go
```diff
@@ -99,22 +100,24 @@ func CreateFromIP(ip net.IP) (Job, error) {
 		return Job{}, err
 	}
 
-	hd := d.Hardware()
-	hwID := hd.HardwareID()
+	if os.Getenv("DATA_MODEL_VERSION") == "1" {
```
mmlb, line 103: How about 
```
if os.Getenv("DATA_MODEL_VERSION != "" {
    return j, nil
}

old new code here
```
instead

mikemrm: I can do that

mmlb: IDK what you're talking about, there's definitely a `!= "1"` up there. All the other uses of DATA_MODEL_VERSION have been in a switch statement and here want to do early return instead.

mikemrm: 🤔  that was definitely something different a second ago, heh. cool, will do.

PR #72, opened 2020-08-26
include syslog host in osie kernel environments

mikemrm:
This allows for a separate host to be defined as the syslog server.
The default action is to fallback to the `PublicIPv4` which is what
is currently used as the DHCP Log Server option value.

You can now define both `PUBLIC_SYSLOG_IP` and `PUBLIC_SYSLOG_FQDN`.
The former is needed to change the dhcp log server. The latter is
passed into the ipxe script.

If the fqdn is not defined, it falls back to the value of the ip.
If the ip is not defined it falls back to the public ip.

### DIFF
--- a/conf/config.go
+++ b/conf/config.go
@@ -14,6 +14,9 @@ var (
 	PublicIPv4 = mustPublicIPv4()
 	PublicFQDN = env.Get("PUBLIC_FQDN", PublicIPv4.String())
 
+	PublicSyslogIPv4 = mustPublicSyslogIPv4()
+	PublicSyslogFQDN = env.Get("PUBLIC_SYSLOG_FQDN", PublicSyslogIPv4.String())
+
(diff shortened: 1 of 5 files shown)


mmlb:
Why not just resolve PUBLIC_SYSLOG_FQDN and use that?

mikemrm:
> Why not just resolve PUBLIC_SYSLOG_FQDN and use that?

🤔  resolving the fqdn sounds dirty. I'd rather not rely on dns internally for something thats mostly cosmetic.

mmlb:
> > Why not just resolve PUBLIC_SYSLOG_FQDN and use that?
> 
> thinking resolving the fqdn sounds dirty. I'd rather not rely on dns internally for something thats mostly cosmetic.

Having 2 vars that sortof mean the same thing feels dirty/weird to me, but meh.

PR #74, opened 2020-08-27
Remove unused MirrorBaseIP variable and associated functions

detiber:
Signed-off-by: Jason DeTiberus <detiber@users.noreply.github.com>

## Description

We were generating a value for the MirrorBaseIP variable that caused issues when specifying a MIRROR_BASE_URL or MIRROR_HOST value that contained a port number, however MirrorBaseIP is never referenced anywhere other than a couple of tests where it is being overridden (to no effect).

## Why is this needed

This allows users to specify MIRROR_BASE_URL or MIRROR_HOST that contains a host:port combination, when the mirror server is not listening on port 80

## How Has This Been Tested?

Tested using the local setup described in the docs, slightly modified to expose the mirror host on a port other than port 80.

This should not affect existing users.

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/conf/mirror.go
+++ b/conf/mirror.go
@@ -2,7 +2,6 @@ package conf
 
 import (
 	"fmt"
-	"net"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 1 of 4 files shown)


detiber:
@mmlb @gianarb ptal

gianarb (approved):

mmlb (approved):

mmlb:
sorry, this dropped off my radar

PR #82, opened 2020-09-16
Create CONTRIBUTING.md

rainleander:
## Description

Our repositories should be the example from which adjacent, competing, projects look for inspiration.

Each repository should not look entirely different from other repositories in the ecosystem, having a different layout, a different testing model, or a different logging model, for example, without reason or recommendation from the subject matter experts from the community.

We should share our improvements with each ecosystem while seeking and respecting the feedback of these communities.

Whether or not strict guidelines have been provided for the project type, our repositories should ensure that the same components are offered across the board. How these components are provided may vary, based on the conventions of the project type. GitHub provides general guidance on this which they have integrated into their user experience.

#75

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,17 @@
+## Hello Contributors!
+  Thx for your interest! We're so glad you're here. 
+
+### Important Resources
+  - bugs: [https://github.com/tinkerbell/boots/issues](https://github.com/tinkerbell/boots/issues)
+
+### Code of Conduct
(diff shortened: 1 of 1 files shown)


mmlb:
@rainleander missed DCO

rainleander:
My B! Fixed.

Signed-off-by: K Rain Leander <rainleander@users.noreply.github.com>

mmlb:
@rainleander this is actually going to go into .github and this file will basically just be a link to that right?

jacobweinstock:
duplicate of https://github.com/tinkerbell/boots/pull/162

PR #84, opened 2020-09-16
Update README.md

rainleander:
## Description

Our repositories should be the example from which adjacent, competing, projects look for inspiration.

Each repository should not look entirely different from other repositories in the ecosystem, having a different layout, a different testing model, or a different logging model, for example, without reason or recommendation from the subject matter experts from the community.

We should share our improvements with each ecosystem while seeking and respecting the feedback of these communities.

Whether or not strict guidelines have been provided for the project type, our repositories should ensure that the same components are offered across the board. How these components are provided may vary, based on the conventions of the project type. GitHub provides general guidance on this which they have integrated into their user experience.

## Why is this needed

Fixes: https://github.com/tinkerbell/boots/issues/85

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,11 @@
 # Boots
 [![Build Status](https://cloud.drone.io/api/badges/tinkerbell/boots/status.svg)](https://cloud.drone.io/tinkerbell/boots)
+![](https://img.shields.io/badge/Stability-Experimental-red.svg)
 
 This services handles DHCP, PXE, tftp, and iPXE for provisions.
+
+This repository is [Experimental](https://github.com/packethost/standards/blob/master/experimental-statement.md) meaning that it's based on untested ideas or techniques and not yet established or finalized or involves a radically new and innovative style!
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,8 +1,11 @@
 # Boots
 [![Build Status](https://cloud.drone.io/api/badges/tinkerbell/boots/status.svg)](https://cloud.drone.io/tinkerbell/boots)
+![](https://img.shields.io/badge/Stability-Maintained-green.svg)
 
 This services handles DHCP, PXE, tftp, and iPXE for provisions.
 
+This repository is [Maintained](https://github.com/packethost/standards/blob/master/maintained-statement.md) meaning that this software is supported by Tinkerbell and its community - available to use in production environments.
```
mmlb, line 7: > available to use in production environments

I'm worried this signals something that is not true.

rainleander: Which part are you concerned about? The Tinkerbell bit or the supported bit or production environment bit or ...?

mmlb (approved):

PR #86, opened 2020-09-29
flatcar: exclude NVMe disks for s3.xl

dustinmiller1337:
## Description

For the s3.xlarge plan type, lets exclude NVMe disks from the installation targets

## Why is this needed

Seems Flatcar isn't working well with those. Also they are marked for cache instead of boot in any case.

## How Has This Been Tested?
Confirmed NVMe only installs seem broken by creating m3.large NVMe only setup.
Barely tested otherwise.

No break, only fix 🤞


mmlb (approved):
:crossed_fingers: no breaksies

PR #88, opened 2020-10-12
flatcar: move to systemd-network

dustinmiller1337:
Currently it seems that Flatcar encounters a race condition on some of 
our hardware. This fix was suggested by @t-lo and crafted by 
@nicolerenee

## Description

Adjust the systemd unit we use to wait for network connectivity

## Why is this needed
Recently Flatcar stopped working on c1 and c3 small.x86 plan types

## How Has This Been Tested?
Highly speculative, untested

No break, only fix 🤞


nicolerenee:
Totally my code with Dustin's name on it, I definitely would treat it as highly suspect. 🤣

mmlb:
LGTM, lets find out if it works together! :D

mmlb (approved):

PR #89, opened 2020-10-12
flatcar: networkd switch from target to service

dustinmiller1337:
Lets point at the service instead of the target. Credit to @nicolerenee

## Description
Fixes a small oversight in #88 where the target is now a service


dustinmiller1337:
Two commits here, let me know if I should break these apart- the build from this seems to resolve the issue 🎉

mmlb (approved):
lgtm this time too!

PR #92, opened 2020-10-23
f3.x86: opt for ttyS0 instead

dustinmiller1337:
Seems these new HP machines use this TTY instead

## Description

For the f3.medium and f3.large plans, lets use ttyS0 instead

## Why is this needed

Not working on ttyS1

### DIFF
--- a/installers/osie/main.go
+++ b/installers/osie/main.go
@@ -122,7 +122,7 @@ func kernelParams(action, state string, j job.Job, s *ipxe.Script) {
 		}
 	} else {
 		s.Args("console=tty0")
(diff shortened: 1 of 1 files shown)


mikemrm (dismissed):

mikemrm (approved):
this should be a case

dustinmiller1337:
```// default
var console string

if j.IsARM() {
  console = "ttyAMA0"
  } else {
  switch planSlug() {
    case  "baremetal_hua",
          "d1p.optane.x86",
          "d1f.optane.x86",
          "f3.medium.x86",
          "f3.large.x86":
      console = "ttyS0"
    default:
      console = "ttyS1"
  }```

is the case I had outlined but wasn't able to get working

PR #94, opened 2020-10-27
Build in docker image with multi platform support

gianarb:


Dockerfile
```diff
@@ -1,8 +1,19 @@
-FROM alpine:3.10
+FROM nixos/nix
```
mmlb, line 1: We should pin to a nix image, we did actually run into build issues a little while ago.

Dockerfile
```diff
@@ -1,8 +1,19 @@
...
 
+RUN nix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs
+RUN nix-channel --update
```
mmlb, line 4: this is unnecessary as we pin nixpkgs in shell.nix

Dockerfile
```diff
@@ -1,8 +1,19 @@
...
+
+COPY . /usr/myapp
+WORKDIR /usr/myapp
+
+RUN nix-shell
+RUN make boots
```
mmlb, line 10: I don't think this works as nix-shell will setup the binaries only for its run and then there's nothing to do so exits and then the next `RUN` doesn't have them.

We'd need something like `RUN nix-shell --run 'make boots'

mmlb (changes requested):

mmlb (approved):

PR #95, opened 2020-10-28
Add discover bootScript

ScottGarman:
This is for DiscoveryOS mode, a new mode of OSIE that offers various
tools to auto-detect and setup hardware during initial rack enrollment
of a server.

### DIFF
--- a/installers/osie/ipxe_script_test.go
+++ b/installers/osie/ipxe_script_test.go
@@ -93,6 +93,18 @@ func TestScript(t *testing.T) {
 }
 
 var prefaces = map[string]string{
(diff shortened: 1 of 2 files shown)


ScottGarman:
Keeping this as a draft until it's received some testing, which I'm still figuring out how to do.

ScottGarman:
I tested this locally today with lots of help from @mikemrm by bringing up cacher and boots containers, pushing some fake data to cacher, and confirming that the normal auto.ipxe file is returned as usual, and that the discover action is set when the operating system is set to discovery. So we're pretty confident this won't introduce any breaking changes, and does what it is intended to do.

mmlb (approved):
lgtm

PR #99, opened 2020-11-02
Cacher2tink

mmlb:
## Description

Create a contrib command to help EM operators transitioning/testing by taking cacher data and transforming into tinkerbell form.

Closes: ENG-10228

### DIFF
--- a/contrib/cacher2tink/main.go
+++ b/contrib/cacher2tink/main.go
@@ -1,0 +1,96 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/tinkerbell/boots/packet"
+)
+
+func main() {
+	var m map[string]interface{}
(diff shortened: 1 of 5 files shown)


mmlb:
@gianarb doesn't look like the update actually fixes the issue :(

gianarb:
Hello! Cacher is not part of the Tinkerbell project and I am not sure how many companies are using it other than Equinix Metal. Do you think this is the right place vs packethost github org? Probably I will prefer to keep it in packethost org

mmlb:
I doubt any companies are using it outside of EM, but regardless of all that EM still uses it very heavily and as we ramp up our efforts to move from legacy cacher -> future tink having this tool in hand will help with testing. This is exactly the type of tool I'd expect to see in a contrib/ directory, but to be honest it would make more sense in tink/ repo over boots but I didn't want to introduce any cacher things into it.

mmlb:
@gianarb actually I think I'm going to move this over to cacher repo instead!

PR #100, opened 2020-11-02
Update golangci lint action

mmlb:
## Description

Hopefully this fixes golangci-lint action spamming CI with unrelated errors. This is both noisy in the stderr output and also ends up failing the action itself. For example:

https://github.com/tinkerbell/boots/runs/1344481754?check_suite_focus=true#step:6:64

> Error: /bin/tar: ../../../go/pkg/mod/github.com/andreyvit/diff@v0.0.0-20170406064948-c7f18ee00883/doc.go: Cannot open: File exists

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,7 +22,7 @@ jobs:
     - name: go vet
       run: go vet ./...
     - name: golangci-lint
-      uses: golangci/golangci-lint-action@v1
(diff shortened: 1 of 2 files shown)


mmlb:
Yep looks like it works!

gianarb:
yes thanks!

gianarb (approved):

mmlb:
@splaspood I choose you!

splaspood (approved):

PR #104, opened 2020-11-16
Eng 8624 os field mismatch

kdeng3849:
Signed-off-by: Kelly Deng <kelly@packet.com>

## Description

This PR fixes the mismatch in field names [`operating_system_version`](https://github.com/tinkerbell/tink/blob/master/protos/packet/packet.proto#L86) vs [`operating_system`](https://github.com/tinkerbell/hegel/blob/a3138d417536903dcdedd674d634e13ebc19fc79/http_handlers.go#L33).

## Why is this needed
This bug mainly affects the users who needs both the `packet.pb.go` and the EC2 endpoint. Because the EC2 endpoint in Hegel uses the field name `operating_system` to be backwards compatible with Kant, users using the `Metadata` object defined in `packet.pb.go` will face a problem with parsing the operating system field since it was defined as `operating_system_version` there. 
A user could have worked around this by specifying two fields `operating_system` and `operating_system_version` with the same value, inside the same piece of hardware, but this introduces some redundancy and just isn't ideal.

Related PR: https://github.com/tinkerbell/tink/pull/275

Fixes: #

## How Has This Been Tested?

Manually tested

N/A

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/packet/models.go
+++ b/packet/models.go
@@ -21,7 +21,7 @@ type BondingMode int
 
 // Discovery interface is the base for cacher and tinkerbell hardware discovery
 type Discovery interface {
-	Instance() *Instance
+	Instance() Instance
 	MAC() net.HardwareAddr
 	Mode() string
 	GetIP(addr net.HardwareAddr) IP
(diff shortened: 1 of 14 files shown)


packet/models.go
```diff
@@ -106,18 +106,61 @@ func NewDiscovery(b []byte) (Discovery, error) {
 }
 
 // Instance models the instance data as returned by the API
-type Instance struct {
+type Instance interface {
```
mmlb, line 109: I don't think this is a good route to take since this is exporting implementation details to callers and making all this more complicated. I think the easiest option is to have an unexported `instance` struct that has both `OperatingSystem` and `OperatingSystemVersion` and then we grab the data from the correct field depending which mode we're running in and then convert to `Instance`.

mmlb (changes requested):

packet/models.go
kdeng3849: That makes sense, but how do I grab the data and convert to Instance? Is `Instance` an empty interface?

mmlb: Copying fields one-by-one. That sucks but the suck is contained into just cacher code that will go away soon anyway.

PR #107, opened 2020-11-24
Docker image does not contain the right binary

gianarb:
## Description

The docker images we built are multi-arch, but for some reason the boots binary inside is x86 for all of them.

## Why is this needed

We need the right binary with the right architecture 

## How Has This Been Tested?

Not tested yet but it will be locally.

The user is currently impacted because Docker images and binaries released as part of sandbox only works for x86.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,7 @@
-FROM nixos/nix:2.3.6
-
-COPY . /usr/myapp
-WORKDIR /usr/myapp
-
-RUN nix-shell --run 'make boots'
-
 FROM alpine:3.11
-
 ENTRYPOINT ["/usr/bin/boots"]
-# ENV GIN_MODE release
 EXPOSE 67 69 80
-
 RUN apk add --update --upgrade --no-cache ca-certificates socat
-COPY --from=0 /usr/myapp/boots /usr/bin/boots
+COPY . /usr/myapp
+RUN cp /usr/myapp/boots-linux-$(uname -m) /usr/bin/boots
+RUN rm -fr /usr/myapp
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -3,5 +3,5 @@ ENTRYPOINT ["/usr/bin/boots"]
 EXPOSE 67 69 80
 RUN apk add --update --upgrade --no-cache ca-certificates socat
 COPY . /usr/myapp
-RUN rm -fv /usr/myapp
 RUN cp /usr/myapp/boots-linux-$(uname -m) /usr/bin/boots
+RUN rm -fr /usr/myapp
```
mmlb, line 7: can you squash this commit into the previous one, no sense mixing the 2.

mmlb (changes requested):

shell.nix
```diff
@@ -1,11 +1,15 @@
...
+in
+{ pkgs ?
+  import
+    (_pkgs.fetchFromGitHub {
+      owner = "NixOS";
+      repo = "nixpkgs";
+      #branch@date: nixpkgs@2020-11-24
+      rev = "6625284c397b44bc9518a5a1567c1b5aae455c08";
+      sha256 = "1w0czzv53sg35gp7sr506facbmzd33jm34p6cg23fb9kz5rf5c89";
+    }) { }
```
mmlb, line 11: this is fine, but what re-formatted this like this? I use `nixfmt` which I now realize should be in here and part of CI.

gianarb: I think i am using nixfmt as well and that's the result.. maybe

mmlb (approved):

markyjackson-taulia (approved):

PR #111, opened 2020-12-15
Added link back to website

mrmrcoleman:
We're linking to the GH repos directly in various marketing content, so we need to direct users back to the main site.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,8 @@
 [![Build Status](https://github.com/tinkerbell/boots/workflows/For%20each%20commit%20and%20PR/badge.svg)](https://github.com/tinkerbell/boots/actions?query=workflow%3A%22For+each+commit+and+PR%22+branch%3Amain)
 ![](https://img.shields.io/badge/Stability-Experimental-red.svg)
 
-This services handles DHCP, PXE, tftp, and iPXE for provisions.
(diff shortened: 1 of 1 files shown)


markyjackson-taulia (approved):

README.md
```diff
@@ -46,6 +46,8 @@ You can use NixOS shell, which will have the Git-LFS, Go and others
 
 `nix-shell`
 
+Note: for mac users, you will need to comment out the line `pkgsCross.aarch64-multiplatform.buildPackages.gcc` in order for the build to work
+
```
jacobweinstock, line 50: FYI, this is not needed anymore as the part of the build that originally needed this now uses docker. ref: https://github.com/tinkerbell/boots/blob/06204de91d2afb509f6ca9422180c56265f34fa9/rules.mk#L97

jacobweinstock (changes requested):

README.md
mmlb: This was brought in from the merges to master. I've now rebased and its all gone.

jacobweinstock (approved):

jacobweinstock (approved):

PR #115, opened 2020-12-23
hasActiveWorkflow should consider workflows in state PENDING and RUNNING

splaspood:
## Description

When checking for the status of a 'job' boots currently only considers PENDING workflows.   This change makes it consider RUNNING workflows to be 'active' as well.

## Why is this needed

During workflow execution we pass status updates back to our central API by way of boots.   As-is, the hasActiveWorkflow function will reject those status updates as it only considers 'PENDING' workflows to be active.

### DIFF
--- a/job/job.go
+++ b/job/job.go
@@ -41,7 +41,7 @@ type Job struct {
 // hwID is the hardware/worker ID corresponding to the MAC
 func hasActiveWorkflow(wcl *tw.WorkflowContextList) (bool, error) {
 	for _, wf := range (*wcl).WorkflowContexts {
-		if wf.CurrentActionState == tw.State_STATE_PENDING {
+		if wf.CurrentActionState == tw.State_STATE_PENDING || wf.CurrentActionState == tw.State_STATE_RUNNING {
 			joblog.With("workflowID", wf.WorkflowId).Info("found active workflow for hardware")
 			return true, nil
 		}
--- a/job/job_test.go
+++ b/job/job_test.go
@@ -183,12 +183,23 @@ func TestHasActiveWorkflow(t *testing.T) {
 		wcl    *tw.WorkflowContextList
 		status bool
 	}{
-		{name: "test active workflow",
+		{name: "test pending workflow",
 			wcl: &tw.WorkflowContextList{
 				WorkflowContexts: []*tw.WorkflowContext{
 					&tw.WorkflowContext{
-						WorkflowId:         "active-fake-workflow-bde9-812726eff314",
+						WorkflowId:         "pending-fake-workflow-bde9-812726eff314",
 						CurrentActionState: 0,
+					},
+				},
+			},
+			status: true,
+		},
+		{name: "test running workflow",
+			wcl: &tw.WorkflowContextList{
+				WorkflowContexts: []*tw.WorkflowContext{
+					&tw.WorkflowContext{
+						WorkflowId:         "running-fake-workflow-bde9-812726eff314",
+						CurrentActionState: 1,
 					},
 				},
 			},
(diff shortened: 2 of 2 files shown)


job/job_test.go
```diff
@@ -194,7 +194,7 @@ func TestHasActiveWorkflow(t *testing.T) {
 			},
 			status: true,
 		},
-		{name: "test inactive workflow",
+		{name: "test running workflow",
 			wcl: &tw.WorkflowContextList{
 				WorkflowContexts: []*tw.WorkflowContext{
 					&tw.WorkflowContext{
```
kdeng3849, line 200: maybe change the workflow ids from `active-fake-workflow-blah` to `pending-fake...`, `running-fake...`, etc. so there's less room for confusion?

kdeng3849 (changes requested):

job/job_test.go
splaspood: good call, I forgot

kdeng3849 (approved):

mmlb (approved):
lgtm

job/job.go
```diff
@@ -41,7 +41,7 @@ type Job struct {
 // hwID is the hardware/worker ID corresponding to the MAC
 func hasActiveWorkflow(wcl *tw.WorkflowContextList) (bool, error) {
 	for _, wf := range (*wcl).WorkflowContexts {
-		if wf.CurrentActionState == tw.State_STATE_PENDING {
+		if wf.CurrentActionState == tw.State_STATE_PENDING || wf.CurrentActionState == tw.State_STATE_RUNNING {
```
Cbkhare, line 44: @splaspood objective of keeping only `Pending` workflow was to only pxe boot if there is a workflow which can be executed. Since a running workflow (already executed workflow) should not be executed again, it should be kept only `Pending`, instead, a new workflow should be created.

Cbkhare (dismissed):

job/job.go
Cbkhare: IMHO this change is not required.

splaspood: We utilize hasActiveWorkflow to determine if we should accept (and pass on) 'status' messages to our central API.   Without this change, all those messages sent during the execution of a workflow are dropped due to the workflow not being considered to be 'active'.

Cbkhare: okay, in that case, I would suggest we should add a separate method to fetch running workflows similar to `hasActiveWorkflow` one, else this can affect pxe boot workflow. what do you think?

splaspood: I wonder if PXE booting when a workflow is RUNNING is really a problem.   Certainly we have no mechanisms in place currently to 'resume' a workflow execution post reboot (although, that sounds like something we'll want to consider/implement for unrelated reasons).   So if a workflow is RUNNING and the machine gets rebooted, it will PXE boot back into tink-worker, which will then likely fail in some manner.   Without this change, instead the machine would be rebooted, fail to PXE, and still fail to complete the RUNNING workflow.

mmlb: IIRC we always envisioned for tink-worker to be pxe booted to even if there's a workflow already runnning. tink-worker executions should not be arbitrarily blocked, especially if workflow actions are idempotent an interrupted action will be able to pick up and re do its bit.

There can be the case of wanting to execute actions from within an installed machine that might have trouble if the machine is always set to pxe boot. But that seems like a fringe issue. There's always kexec option and possibly using pbnj to ensure pxe boot is disabled.

jgavinray (approved):
LGTM

markyjackson-taulia (approved):
/lgtm

job/job.go
mmlb: @Cbkhare going to dismiss this. This is the way we want to go in general. If you see a case where we don't want to netboot please open up a separate issue.

Cbkhare:
@mmlb / @splaspood I have opened a bug with reference to this commit https://github.com/tinkerbell/boots/issues/125.

Issue #116, opened 2021-01-04
Remove binaries from the project


displague:
## Expected Behaviour

Project binaries should be external artifacts, built through automation, not included in the git repository.

## Current Behaviour

Binaries are included in the project and are expected to be compiled and updated by contributors. The binaries must then be manually verified through recompiling (with the same tools) and checksum verification.  This slows our ability to accept new changes (#79).

## Possible Solution

Similar to #4, the ipxe/bin artifacts should be published to GitHub if they are needed outside of a boots container. If a boots container image is sufficient, the binaries should be compiled at image build time, not trusted from the repository.

The LFS ipxe/bin artifacts should then be removed and the instructions (https://github.com/tinkerbell/boots#local-setup) updated accordingly.

geekgonecrazy:
This would help a lot.  A few other changes I’d like to contribute here on this code base but that part made this component one I drag my feet wanting to even touch

mmlb:
Yep I think this will definitely help getting things upstream, my main hangup is that then this code base then loses its `go get`edness.

mmlb:
:fist_right: lfs, https://github.com/tinkerbell/boots/pull/79#issuecomment-761007091. I guess just one more reason for this then :D.

PR #79, opened 2020-09-16
Add option for snponly when architecture matches

geekgonecrazy:
## Description
Every source online I can find when it comes to ipxe and booting architecture type 00007 ( aka x64 EFI ) suggests using snp.  

In my case snp is the only one that would work.  Since using snp and this is obviously chainloading made sense to use snponly.

So if the architecture matches this serves snponly.efi

## Why is this needed

Some nics are unsupported.  #78 is a good reference of the conversation leading up to this

Fixes: #

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required)
- [ ] added unit or e2e tests
- [ ] provided instructions on how to upgrade

PR #117, opened 2021-01-07
Uncomplicate ipxe build procedure

mmlb:
## Description

This PR gets rid of the multiple makefiles / recursive makes approach used to build iPXE. Simplifying the iPXE build description quite a bit.

## Why is this needed

Updating iPXE build config is quite convoluted and makes contributions harder than necessary. As @geekgonecrazy experienced in #79. This is the first steps to possibly addressing #116. 

## How Has This Been Tested?

`make`

Should be easier to mess with iPXE config/builds.


mmlb:
Hey @displague pinged you for review as you seem interested in this area :D, what with #116 and all.

Makefile
```diff
@@ -1,30 +1,65 @@
...
 
-crosscompile: $(shell git ls-files | grep -v -e vendor -e '_test.go' | grep '.go$$' )
-	CGO_ENABLED=0 GOOS=linux GOARCH=386 go build -v -o ./boots-linux-x86_64 -ldflags="-X main.GitRev=$(shell git rev-parse --short HEAD)"
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -o ./boots-linux-amd64 -ldflags="-X main.GitRev=$(shell git rev-parse --short HEAD)"
-	CGO_ENABLED=0 GOOS=linux GOARCH=arm GOARM=6 go build -v -o ./boots-linux-aarch64 -ldflags="-X main.GitRev=$(shell git rev-parse --short HEAD)"
-	CGO_ENABLED=0 GOOS=linux GOARCH=arm GOARM=7 go build -v -o ./boots-linux-armv7l -ldflags="-X main.GitRev=$(shell git rev-parse --short HEAD)"
-	CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -v -o ./boots-linux-arm64 -ldflags="-X main.GitRev=$(shell git rev-parse --short HEAD)"
-
+GitRev := $(shell git rev-parse --short HEAD)
+crosscompile: ${binary}
```
gianarb, line 19: I have to validate how this will break the release process in sandbox. It will do it for sure, and we should probably document and align the other project and binaries as well otherwise it will be a pretty messy experience.

gianarb (commented):
I don't have much knowledge about what is going on here technically but I like that it simplifies build and it is in preparation for removing checked-in binaries.

As I wrote, we need doc and alignment in terms of binary prefix name/arch.

Makefile
mmlb: This won't change any of that, at least that was the intent.

mmlb: ohhh thats a mistake, not meant for this PR! Let me drop that!

gianarb: oh you are right! Git diff tricked me out!

mmlb: ok yes I got tricked too :D

mmlb: Nothing has changed here :D

mmlb:
Had to do a amend/force push to kick the DCO bot, it looked to be stuck.

gianarb (approved):

PR #122, opened 2021-01-20
Fix binary names

mmlb:
## Description

This PR fixes the binary names to be both correct and also "unsurprising".

## Why is this needed

Avoids needing to COPY all the binaries, only to keep 1. Also avoids glaringly misnamed binaries :D. 

For example these are the binaries built from master:
```
file boots-* | sed 's|, statically.*||'
boots-linux-aarch64: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV)
boots-linux-amd64:   ELF 64-bit LSB executable, x86-64, version 1 (SYSV)
boots-linux-arm64:   ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV)
boots-linux-armv7l:  ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV)
boots-linux-x86_64:  ELF 32-bit LSB executable, Intel 80386, version 1 (SYSV)
```
notice the 32bit arm v6 binary is named aarch64 and x86_64 is actually 386!

Here is from this PR:
```
[~/go/src/github.com/tinkerbell/boots]─[manny@dellnix]> 
file boots-* | sed 's|, statically.*||'
boots-linux-386:   ELF 32-bit LSB executable, Intel 80386, version 1 (SYSV)
boots-linux-amd64: ELF 64-bit LSB executable, x86-64, version 1 (SYSV)
boots-linux-arm64: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV)
boots-linux-armv6: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV)
boots-linux-armv7: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV)
```

## How Has This Been Tested?

Inspecting each file with `file`, docker build works.

Any user using the cross compiled binaries directly will be impacted, but this should most likely be 0 users since they will likely use plain-old `boots` binary instead.


gianarb:
Thank you for making it right! I have to admit, we need a bit of consistency. 
Can you keep the same binary name we already use in the other project? have a look here.
https://github.com/tinkerbell/sandbox/releases/tag/v0.3.0

Otherwise, can you apply the same criteria to the other projects?

mmlb:
@gianarb I think those names in sandbox should be changed to -armv6 and armv7 as that seems to be the most common scheme in go projects I can find releases for (caddy, traefik, golangci-lint) and also has the easiest scheme $name-$os-$cpu

mmlb:
Also, @gianarb only boots does cross compiling outside of docker so there isn't really anything to do for any of the other projects right?

gianarb:
I see what you are saying.. Yep right

gianarb (approved):

PR #124, opened 2021-01-28
ipxe: Do not vendor binaries in git lfs

### LINKED ISSUE #116: Remove binaries from the project

mmlb:
## Description

Removes the iPXE binaries we can actually build out of GitLFS.

## Why is this needed

This makes it easier for contributors to contribute since we
don't have to reproduce or over write the ixpe builds they
contribute. This does mean the project isn't `go get`able but
I guess we'll just have to live with that.

Closes #116

## How Has This Been Tested?

N/A

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,1 @@
-ipxe/bin/ipxe.efi filter=lfs diff=lfs merge=lfs -text
 ipxe/bin/snp-hua.efi filter=lfs diff=lfs merge=lfs -text
-ipxe/bin/snp-nolacp.efi filter=lfs diff=lfs merge=lfs -text
-ipxe/bin/undionly.kpxe filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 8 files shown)


gianarb (changes requested):
Amazing! Thank you for that bu apparently CI is not happy, `go vet` in particular.
Do you mind do have a look?

gianarb (approved):

Issue #114, opened 2020-12-22
Boots should pass Osie to registered hardware without workflow yet


gianarb:
## Expected Behaviour
A registered hardware without a workflow registered should boot into Osie.

## Current Behaviour
Boots logs:

```
{"level":"info","ts":1608652334.4974012,"caller":"boots/dhcp.go:76","msg":"retrieved job is empty","service":"github.com/tinkerbell/boots","pkg":"main","type":"DHCPDISCOVER","mac":"f4:4d:30:64:8e:0f","err":"discover from dhcp message: get hardware by mac from tink: rpc error: code = Unknown desc = unexpected end of JSON input","errVerbose":"rpc error: code = Unknown desc = unexpected end of JSON input\nget hardware by mac from tink\ngithub.com/tinkerbell/boots/packet.(*Client).DiscoverHardwareFromDHCP\n\t/home/runner/work/boots/boots/packet/endpoints.go:104\ngithub.com/tinkerbell/boots/job.discoverHardwareFromDHCP.func1\n\t/home/runner/work/boots/boots/job/fetch.go:16\ngithub.com/golang/groupcache/singleflight.(*Group).Do\n\t/home/runner/go/pkg/mod/github.com/golang/groupcache@v0.0.0-20190702054246-869f871628b6/singleflight/singleflight.go:56\ngithub.com/tinkerbell/boots/job.discoverHardwareFromDHCP\n\t/home/runner/work/boots/boots/job/fetch.go:18\ngithub.com/tinkerbell/boots/job.CreateFromDHCP\n\t/home/runner/work/boots/boots/job/job.go:59\nmain.dhcpHandler.serveDHCP\n\t/home/runner/work/boots/boots/dhcp.go:74\nmain.dhcpHandler.ServeDHCP.func1\n\t/home/runner/work/boots/boots/dhcp.go:45\ngithub.com/gammazero/workerpool.startWorker\n\t/home/runner/go/pkg/mod/github.com/gammazero/workerpool@v0.0.0-20200311205957-7b00833861c6/workerpool.go:218\nruntime.goexit\n\t/opt/hostedtoolcache/go/1.15.5/x64/src/runtime/asm_386.s:1333\ndiscover from dhcp message"}
```
And it does not serve Osie via iPXE

## Possible Solution

Serve osie even if there are not workflow registered.

## Steps to Reproduce (for bugs)
1. Start tinkerbell with Vagrant or what ever
2. Register hardware
3. Start the hardware without any workflow

[...]

PR #133, opened 2021-03-11
Allow PXE booting regardless of an existing workflow:

jacobweinstock:
## Description

Currently, a machine will only PXE boot if it has an existing tink workflow.
To do any other type of PXE boot (custom ipxe, osie, hook, etc) you have to create a workflow that will never actually be run.
This commit allows machines to PXE into OSIE, hook, OSIE alternative, custom ipxe, etc without needing a workflow.
PXE booting will be gated by the tink hardware data field: `network.interfaces[].netboot.allow_pxe`

## Why is this needed

Allows machines to PXE boot without a workflow.

Fixes: https://github.com/tinkerbell/boots/issues/114

## How Has This Been Tested?

I built the boots image and ran it with the vagrant sandbox locally to test. It worked as intended there. 

no migrations.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/http.go
+++ b/http.go
@@ -117,16 +117,15 @@ func serveJobFile(w http.ResponseWriter, req *http.Request) {
 		mainlog.With("client", req.RemoteAddr, "error", err).Info("no job found for client address")
 		return
 	}
-
-	activeWorkflows, err := job.HasActiveWorkflow(j.HardwareID())
-	if err != nil {
+	// This gates serving PXE file by
+	// 1. the existence of a hardware record in tink server
+	// AND
+	// 2. the network.interfaces[].netboot.allow_pxe value, in the tink server hardware record, equal to true
+	// This allows serving custom ipxe scripts, starting up into OSIE or other installation environments
+	// without a tink workflow present.
+	if !j.AllowPxe() {
 		w.WriteHeader(http.StatusNotFound)
-		mainlog.With("client", req.RemoteAddr, "error", err).Info("failed to get workflows")
-		return
-	}
-	if !activeWorkflows {
-		w.WriteHeader(http.StatusNotFound)
-		mainlog.With("client", req.RemoteAddr).Info("no active workflows")
+		mainlog.With("client", req.RemoteAddr).Info("the hardware data for this machine, or lack there of, does not allow it to pxe; allow_pxe: false")
 		return
 	}
 
--- a/job/job.go
+++ b/job/job.go
@@ -36,6 +36,12 @@ type Job struct {
 	instance *packet.Instance
 }
 
+// AllowPxe returns the value from the hardware data
+// in tink server defined at network.interfaces[].netboot.allow_pxe
+func (j Job) AllowPxe() bool {
+	return j.hardware.HardwareAllowPXE(j.mac)
+}
+
 // HasActiveWorkflow fetches workflows for the given hardware and returns
 // the status true if there is a pending (active) workflow
 func HasActiveWorkflow(hwID packet.HardwareID) (bool, error) {
--- a/tftp.go
+++ b/tftp.go
@@ -58,14 +58,15 @@ func (tftpHandler) ReadFile(c tftp.Conn, filename string) (tftp.ReadCloser, erro
 		return serveFakeReader(l, filename)
 	}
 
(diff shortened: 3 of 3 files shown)


gianarb (approved):
Thank you for taking care of this!!

gianarb:
Now that I am thinking about it @jacobweinstock do you mind to figure out and add tests for this? not sure about the status of the test suite but @mmlb knows more I suppose

mmlb:
Hmm the issue with this atm is that there's nothing to turn it off once a workflow is done. This is likely to be a problem when machines are set to boot from nic by default, which is probably the case for NUCs/machines w/o a BMC, or even servers with PXE boot as default because of fw issues (some EM machines are like this)

markyjackson-taulia (approved):

mmlb:
We don't really have a good story for messing with hw data. Should tink-(worker/server) disable allow_pxe if the workflow is finished? Is this another case for "builtin tink-worker actions"? Do we build something via ephemeral_data to signal this?

jacobweinstock:
> Hmm the issue with this atm is that there's nothing to turn it off once a workflow is done. This is likely to be a problem when machines are set to boot from nic by default, which is probably the case for NUCs/machines w/o a BMC, or even servers with PXE boot as default because of fw issues (some EM machines are like this)

Setting this stops boots from handing out pxe stuff.
network.interfaces[].netboot.allow_pxe=false

Also at the moment, a provision is only started when some external source sets a machine to pxe boot and restarts it. This approach relies on this same idea. If a machine shouldn't be pxe booted then it shouldn't set itself to pxe boot.

Whether tink controls the setting of boot devices and power cycling or not doesn't feel like it should matter. If a machine is set to pxe, boots should pxe.

What is pxe'd is a bit different. Is it a discovery OS, OSIE and a workflow, a custom ipxe script, a live CD? I feel that what is pxe'd will most likely be very specific to the environment where tink is deployed. Those business rules should be flexible and accommodate all users. Encoding those in boots doesn't feel like the correct approach.

The change at least allows machines to not have to create useless workflows just to pxe. It creates more flexibility by giving more control back to the environment.

gianarb:
I agree with @jacobweinstock here. We have a field already set that pilots if a machine should boot or not via PXE boot `network.interfaces[].netboot.allow_pxe=false`. The fact that you have to create a workflow is not documented and counterintuitive.

> Hmm the issue with this atm is that there's nothing to turn it off once a workflow is done. This is likely to be a problem when machines are set to boot from nic by default, which is probably the case for NUCs/machines w/o a BMC

I think the power cycle issue should not prevent us to make this change. It is an issue on its own. For example, I have 10 NUC and I have recently figured out how I can hook an RPI to them to pilot the power lifecycle.

We have also the ongoing effort to support reboot, power off in tink-worker, and hook. So, it is an issue on its own.

mmlb:
SGTM. Internally we have a separate service that will update the allow_pxe field so we'd be fine. I wanted to make sure we don't make it harder for other members of the community to mess with tinkerbell.

> I agree with @jacobweinstock here. We have a field already set that pilots if a machine should boot or not via PXE boot `network.interfaces[].netboot.allow_pxe=false`. The fact that you have to create a workflow is not documented and counterintuitive. I opened this issue because during my experiments with NUCs I found that not an ideal behaviour.
> 
Also agree with this.

> > Hmm the issue with this atm is that there's nothing to turn it off once a workflow is done. This is likely to be a problem when machines are set to boot from nic by default, which is probably the case for NUCs/machines w/o a BMC
> 
> I think the power cycle issue should not prevent us to make this change. It is an issue on its own. For example, I have 10 NUC and I have recently figured out how I can hook an RPI to them to pilot the power lifecycle.

But that doesn't address what I brought up. Your NUCs are set to PXE boot by default right? Your RPi doesn't configure that on them does it?

I think we should definitely merge this but I'm also trying to figure out if we can have a nice way to not have to leave it up to docs that you'd want to go back and edit the hw and change the allow_pxe: false after done with a workflow.

mmlb:
Maybe we break this off into a discussion for the last bit I mentioned.

mmlb (approved):

PR #135, opened 2021-03-12
support the provisioner_engine metadata value

splaspood:
## Description

Boots should check the provisioiner_engine metadata value against the PROVISIONER_ENGINE_NAME runtime environment to decide if it should be handling DHCP requests for a given piece of hardware.

## Why is this needed

To allow for multiple differing versions of boots to be running in an environment, with DHCP relayed to both simultaneously to aid in blue/green deployment testing.

### DIFF
--- a/job/dhcp.go
+++ b/job/dhcp.go
@@ -90,7 +90,11 @@ func (j Job) isPXEAllowed() bool {
 }
 
 func (j Job) areWeProvisioner() bool {
-	return j.hardware.HardwareProvisioner() == j.ProvisioningEngineName()
+	if len(j.hardware.HardwareProvisioner()) > 0 {
+		return j.hardware.HardwareProvisioner() == j.ProvisionerEngineName()
+	}
+
+	return true
 }
 
 func (j Job) setPXEFilename(rep *dhcp4.Packet, isPacket, isARM, isUEFI bool) {
--- a/job/dhcp.go
+++ b/job/dhcp.go
@@ -90,7 +90,7 @@ func (j Job) isPXEAllowed() bool {
 }
 
 func (j Job) areWeProvisioner() bool {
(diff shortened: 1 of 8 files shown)


job/dhcp.go
```diff
@@ -28,6 +28,13 @@ func IsSpecialOS(i *packet.Instance) bool {
 // ServeDHCP responds to DHCP packets
 func (j Job) ServeDHCP(w dhcp4.ReplyWriter, req *dhcp4.Packet) bool {
 
+	// If we are not the chosen provisioner for this piece of hardware
+	// do not respond to the DHCP request
+	if !j.areWeProvisioner() {
+		j.Error(errors.New("not the provisioner for this piece of hardware"))
```
mmlb, line 34: Should be `j.With("error", errors.New(...)).Info()`

mmlb: I'm not sure we really care to log this though? This is going to be spammy right?

mmlb (changes requested):

job/dhcp.go
splaspood: Yeah I figured we might end up not logging this at all as it will be majorly spammy...   but we already are majorly spammy, so I thought maybe... ;).     I'll just ditch the logging.

job/dhcp.go
```diff
@@ -83,6 +89,10 @@ func (j Job) isPXEAllowed() bool {
 	return j.instance.AllowPXE
 }
 
+func (j Job) areWeProvisioner() bool {
+	return j.hardware.HardwareProvisioner() == j.ProvisioningEngineName()
```
mmlb, line 93: Would it make sense to return true if its not set? This way we don't *require* this setting?

mmlb (approved):

PR #139, opened 2021-03-25
Fix panic from syslog trying to log a nil error with Error

mmlb:
## Description

The syslog package was logging severities 0-3 (emerg, alert, crit, err) using `log.Error`. This ends up passing a nil error to `Error()` which `pkg/log`'s rollbar tries to call `.Error` on. And boom panic.

This is PR handles this by not logging as `Error` in the syslog package, as its not really a boots error anyway, and by updating to version of `packethost/pkg/log` that returns from `Error` if error is nil.

## How Has This Been Tested?

Ran locally in dev testing. Will be tried in lab/staging.

No panics due to syslog messages.


markyjackson-taulia (approved):

syslog/receiver.go
```diff
@@ -11,9 +11,17 @@ import (
 )
 
 var (
-	syslogAddr = conf.SyslogBind
-
-	loggerFuncs map[byte]func(...interface{})
+	syslogAddr  = conf.SyslogBind
+	loggerFuncs = map[byte]func(...interface{}){
```
jacobweinstock, line 15: this appears to only be used at line 122. maybe don't need a package variable https://github.com/tinkerbell/boots/blob/e58c69257db494a216b41813b27f0d230dd195c6/syslog/receiver.go#L122

Makefile
```diff
@@ -12,31 +12,32 @@ SHELL := bash
 .SHELLFLAGS := -o pipefail -euc
 
 binary := boots
```
jacobweinstock, line 14: looks like this isn't used anymore, might consider removing it

Makefile
```diff
@@ -66,36 +67,24 @@ ipxe/ipxe/build/bin-arm64-efi/snp.efi ipxe/ipxe/build/bin-x86_64-efi/ipxe.efi ip
...
+golangci-lint: ## Run golangci-lint 
 	@echo be sure golangci-lint is installed: https://golangci-lint.run/usage/install/
 	golangci-lint run
 
-.PHONY: validate-local
-validate-local: vet go-test goimports golangci-lint # validate-local runs all the same validations and tests that CI run
-	
+validate-local: vet coverage goimports golangci-lint ## Runs all the same validations and tests that run in CI
 	
+help: ## Print this help
```
jacobweinstock, line 89: yeah!!

jacobweinstock (dismissed):
couple of comments but nothing to block on

Makefile
mmlb: yep thanks find/replace doesn't make unnecessary things go away :D

Makefile
mmlb: :highfive:

syslog/receiver.go
mmlb: very true, will drop

mmlb:
@jacobweinstock good call on the `loggerFuncs`, I uh couldn't help myself and needed to make the other maps go away too, PTAL.

jacobweinstock (approved):

jacobweinstock (approved):
do we need a `go generate` call in the `Makefile` somewhere now?

markyjackson-taulia (approved):

markyjackson-taulia:
Merging in known borked state and PR will following to add `go generate` to makefile

PR #141, opened 2021-03-26
http: Only check for HasActiveWorkflow if CanWorkflow is true

mmlb:
## Description

Only checks for workflows if we're in tinkerbell mode and setup for workflows.

## Why is this needed

Would break EM provisions for legacy sites.

## How Has This Been Tested?

Locally via docker-compose.

EM users get a fancy new boots release.

### DIFF
--- a/http.go
+++ b/http.go
@@ -147,16 +147,18 @@ func serveHardware(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


markyjackson-taulia (approved):
aha, a change!  Nice!

markyjackson-taulia (dismissed):

splaspood (approved):

PR #147, opened 2021-04-16
Refactor all table driven tests to use t.Run

mmlb:
## Description

Use `t.Run` in all the table driven tests.

## Why is this needed

Using `t.Run` gives us better selectivity with `go test -run` and also better grouping/splitting when running with `go test -v`.

## How Has This Been Tested?

Just minor test changes, so tested with `go test`.


mmlb:
This actually depends on #146

markyjackson-taulia (approved):

markyjackson-taulia (approved):

PR #148, opened 2021-04-20
ci: Run builds on our self hosted builders

mmlb:
## Description

Run ci builds on our self hosted, bare metal runners.

## Why is this needed

Faster builds/results especially for the iPXE builds.

## How Has This Been Tested?

GitHub Actions

Faster test cycles

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,12 +20,6 @@ jobs:
         uses: cachix/install-nix-action@018abf956a0a15673dae4932ae26f0f071ac0944
         with:
           nix_path: nixpkgs=channel:nixpkgs-unstable
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
-        with:
-          version: "v0.4.1"
       - name: Non Go formatters and linters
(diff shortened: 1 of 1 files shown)


mmlb:
depends on getting git-lfs in the runners, either via https://github.com/tinkerbell/infrastructure/pull/9 or by installing locally.

tstromberg:
The `Install nix` test is failing at `checkExecSyncError` - can you get this cleared up?

mmlb:
Yep, I believe we now have builders with nix preinstalled so no need for that bit.

mmlb:
looks like self-hosted runners now have nix :raised_hands: so will need to mess with this to work around some of the self-hosted GHA differences.

mmlb:
@gianarb/@tstromberg ready for re-review!

matoszz (approved):
a slight nit, but I would suggest splitting out the go related tasks into it's own actions workflow so that you only run it when the go files are modified, otherwise you're doing a lot of setup + re-running steps unnecessarily

mmlb:
> a slight nit, but I would suggest splitting out the go related tasks into it's own actions workflow so that you only run it when the go files are modified, otherwise you're doing a lot of setup + re-running steps unnecessarily

meh idk if its worth it since most of the time the go files are in fact being touched.

PR #149, opened 2021-04-23
Refactor setPXEFilename to be less complicated

mmlb:
## Description

Pretty large clean up of job.setPXEFilename to get rid of (some) Packet/EM-isms and clean up the logic around sending a file name or not.

## Why is this needed

I'm working on removing cacher and this test was getting in the way, it was too complicated to keep.

## How Has This Been Tested?

Unit tests pass. Still need to do a few boot/provision tests.

NA

## Checklist:

I have:

- [x] added unit or e2e tests


job/hardware.go
```diff
@@ -27,28 +27,28 @@ type ComponentsResponse struct {
 func (j Job) AddHardware(w http.ResponseWriter, req *http.Request) {
 	b, err := readClose(req.Body)
 	if err != nil {
-		joblog.Error(errors.Wrap(err, "reading hardware component body"))
+		j.Error(errors.Wrap(err, "reading hardware component body"))
```
jacobweinstock, line 30: what do you think about using `j.Logger.Error` here? really helps make it clear that this is a log line.

markyjackson-taulia (approved):
Approved pending successful CI run

ScottGarman (dismissed):

job/hardware.go
mmlb: :thinking: yeah that makes more sense. `j.Error` doesn't actually log!

mmlb: fixed in 0b0614cabc013d3990d09d0a78022922a1685020

job/dhcp.go
```diff
@@ -97,64 +111,31 @@ func (j Job) areWeProvisioner() bool {
...
-		if j.instance.State != "active" {
-			j.With("hardware.state", j.HardwareState(), "instance.state", j.instance.State).Info("device should NOT be trying to PXE boot")
-			return
+func (j Job) getPXEFilename(arch, firmware string, isOuriPXE bool) string {
+	if !j.isPXEAllowed() {
+		if j.instance != nil && j.instance.State == "active" {
+			// We set a filename because if a machine is actually trying to PXE and nothing is sent it may hang for
+			// a while waiting for any possible ProxyDHCP packets and it would delay booting from disks.
+			// This short cuts all that when we know we want to be booting from disk.
+			return "/pxe-is-not-allowed"
```
jacobweinstock, line 120: should this logic be in `isPXEAllowed()`? if we're defining when a machine should pxe or not might be good to have it all in one place. that way when we remove it it's all together! 😁

splaspood: On an unrelated note, I've wondered if rather than returning this /pxe-is-not-allowed we should instead actually return a proper bootloader, that simply boots from HD.   There is an example of this in the netboot.xyz.    I wonder if that would be clearer to the end user.

mmlb: both good points. I'll fix @jacobweinstock's in this PR and open up an issue for your @splaspood

mmlb: oh wait no @jacobweinstock no this doesn't belong in `isPXEAllowed` as that doesn't return a filename and we want to differentiate between "boot to disk by returning a bogus name" vs "really should not be pxe'ing and we're better off causing the boot process to hang, so return nothing"

mmlb:
@jacobweinstock / @splaspood rebased on top of latest master, no other changes.

job/dhcp.go
jacobweinstock: So my first comment didn't capture the root concern that I have. Apologies. What follows here are all suggestions, no worries at all if you disagree. We don't have any written design philosophies so this is all just my opinion. I don't see any reason why this code won't work as-is. I will not hold up this PR if you don't want to make any of the changes I'm suggesting.

My main concern is that I tried to create a mental model/flow chart of the logic for whether to PXE or not and then what to provide if we should PXE but I was unable to do that without having to walk through every line of code starting from `if dhcp.SetupPXE(rep, req) {`.

I found that the naming of functions did not accurately describe their purpose/function. Most of them were multi-purpose functions. With duplicate behavior splattered across many of them. I can provide more details about this if requested.

With that said here are some suggestions just related for `getPXEFilename`. All this with the goals of increasing understandability and maintainability, and decreasing fragility:

- remove `isOuriPXE bool` from the `getPXEFilename `signature
  - if the caller already knows this value is true they should just use `"http://" + conf.PublicFQDN + "/auto.ipxe"`
  - ```go
    if isOuriPXE {
        dhcp.SetFilename(rep, "http://" + conf.PublicFQDN + "/auto.ipxe", conf.PublicIPv4)
    }

    ```
- redefine `getPXEFilename` to just determine a PXE filename based on its inputs. This increases the understandability of the function.
- with a func name of`getPXEFilename` we should define through func comments what this means. something like
```go
// getPXEFilename takes in an architecture and firmware type to determine the filename that should be returned
func (j Job) getPXEFilename(arch, firmware string) string {
``` 
- move and rename `j.isPXEAllowed()` to live in `configureDHCP`. Use the positive if case, something like
```go
if j.bootToDiskImmedately() && j.instance != nil && j.instance.State == "active"{
    dhcp.SetFilename(rep, "/pxe-is-not-allowed", conf.PublicIPv4)
} else {
    dhcp.SetFilename(rep, "", conf.PublicIPv4)
}
```


Ref [Curly's Law](https://blog.codinghorror.com/curlys-law-do-one-thing/#:~:text=Curly's%20Law%20is%20about%20choosing,your%20code%20won't%20do.)

jacobweinstock (commented):
Would adding some tests be a possibility, by chance?

job/dhcp.go
jacobweinstock: sorry, one last thing. You said
> _"boot to disk by returning a bogus name" vs "really should not be pxe'ing and we're better off causing the boot process to hang, so return nothing"_

this business logic feels very opaque. I really would like to see us rethink and reorganize this codebase to be clearer and if possible less opinionated about this type of logic. Anyway, just something to think about.

mmlb: > sorry, one last thing. You said

No worries, keep em coming!

> 
> > _"boot to disk by returning a bogus name" vs "really should not be pxe'ing and we're better off causing the boot process to hang, so return nothing"_
> 
> this business logic feels very opaque. I really would like to see us rethink and reorganize this codebase to be clearer and if possible less opinionated about this type of logic. Anyway, just something to think about.

Agree, I hope by removing cacher stuff we can then make progress on removing packet/EM only things (phone-home, password callback, ...) and then have a much smaller code base where we can see all this opinionated business logic that we can hash out if they make sense or not and clean up if they do.

mmlb:
@jacobweinstock I can't seem to reply to your minimized comment, so here goes. Many good points and I almost did some of that in this PR but backed off from doing so to avoid making it even larger or wider in scope. I wanted to try and keep the core of the PR to cleaning up the function and then later can come back with some other refactorings.

jacobweinstock (approved):

ScottGarman:
How do we want to deal with the code coverage CI failures? I'm reviewing some stale PRs and I don't think we want to let this one die on the vine, it's really close, though it needs a rebase too.

splaspood:
It's been long enough that this PR should be rebased.    I'd argue we should make the code coverage check happy, or we should get rid of it.

tstromberg:
@mmlb - This PR was approved. Can you resolve the merge conflicts?

ipxe/dhcp_options.go
```diff
@@ -68,16 +68,16 @@ func Setup(rep *dhcp4.Packet) {
 	rep.SetIP(dhcp4.OptionLogServer, conf.PublicSyslogIPv4) // Have iPXE send syslog to me.
 }
 
-var packetVersion = []byte{1, 0, 255}
+var ouriPXEVersion = []byte{1, 0, 255}
```
jacobweinstock, line 71: what does this correspond to? version 255? how is that our ipxe version?

dhcp/pxe.go
```diff
@@ -123,16 +123,13 @@ func SetupPXE(ctx context.Context, rep, req *dhcp4.Packet) bool {
...
-func SetFilename(rep *dhcp4.Packet, filename string, nextServer net.IP, pxeClient bool) {
+func SetFilename(rep *dhcp4.Packet, filename string, nextServer net.IP) {
 	file := rep.File()
 	if len(filename) > len(file) {
 		err := errors.New("filename too long, would be truncated")
 		// req CHaddr and XID == req's
 		dhcplog.With("mac", rep.GetCHAddr(), "xid", rep.GetXID(), "filename", filename).Fatal(err)
 	}
-	if pxeClient {
-		rep.SetString(dhcp4.OptionClassID, "PXEClient")
```
jacobweinstock, line 134: why is this being removed? im pretty sure we need to respond to pxe clients with this. is it being set elsewhere, maybe?

ref: Table 2-3 and Table 2-5 of http://www.pix.net/software/pxeboot/archive/pxespec.pdf

job/dhcp.go
```diff
@@ -70,23 +70,43 @@ func (j Job) configureDHCP(ctx context.Context, rep, req *dhcp4.Packet) bool {
...
+		arch := "x86"
+		firmware := "bios"
+
 		isARM := dhcp.IsARM(req)
 		if dhcp.Arch(req) != j.Arch() {
 			span.AddEvent(fmt.Sprintf("arch mismatch: got %q and expected %q", dhcp.Arch(req), j.Arch()))
 			j.With("dhcp", dhcp.Arch(req), "job", j.Arch()).Info("arch mismatch, using dhcp")
 		}
+		if isARM {
+			arch = "arm"
```
jacobweinstock, line 82: might be useful to make these values (`arm`, `hua`, `2a2`, etc) constants?

jacobweinstock (changes requested):

job/dhcp.go
mmlb: I'm not too convinced that going from bare strings to a token is very helpful in this case. `arch` isn't going to be const, and I think that 

```go
arch = "arm"
...
case arch == "arm"
```
just reads better than
```go
arch = archARM
...
case arch == archARM
```.

The arch is strings are used in just 2 locations (setting/testing) and I don't think would be worth it to have an indentifier to avoid a little repitition.

File: dhcp/pxe.go (hunk 4 above)
Comment (mmlb) [reply dhcp/pxe.go:134]: great catch, I was thinking the wrong thing here. Will drop.

File: ipxe/dhcp_options.go (hunk 3 above)
Comment (mmlb) [reply ipxe/dhcp_options.go:71]: We override the iPXE version that iPXE uses for itself and sends to the dhcp server. This is how we distinguish an iPXE burned into a nic by a vendor vs the one we build and send via tftp. There are better options here, but not as convenient to apply. We can define the IPXE_VERSION via an env var when building ipxe but other options would require a patch to one or more locations in the iPXE code. This version number is no longer incrementing in upstream iPXE so is low-risk to modify.

File: dhcp/pxe.go (hunk 4 above)
Comment (mmlb) [reply dhcp/pxe.go:134]: Actually going from the code I removed it is *not* needed. It was only being set to true when the incoming DHCP request is from *our* iPXE version. `pxeClient` is never set to true in the `!isPacket` branch in the [code](https://github.com/tinkerbell/boots/blob/6f892cb13e7643e9623d278196984d57f31b6e5b/job/dhcp.go#L142-L152). That means that the NIC was happy to boot the tftp file we sent it w/o the said PXEClient ClassID. I still think we should set it anyway though, but I'll move it to `SetupPXE` instead of setting it depending on the iPXE version or what not.

Comment (mmlb) [reply dhcp/pxe.go:134]: `Section 2.2.2: Protocol Timeouts` 2nd paragraph states:
> The purpose of the timeouts is to ensure the PXE client gives precedence to servers supplying
“PXEClient” specific configuration tags. The PXE boot ROM must function as a normal DHCP boot
ROM in the absence of a PXE specific response. However, the PXE boot ROM must wait for
specified times to see if a PXE response is available before using a non-PXE configuration.

So having it is actually not *required* but a very good thing to have anyway.

File: ipxe/dhcp_options.go (hunk 3 above)
Comment (jacobweinstock) [reply ipxe/dhcp_options.go:71]: I believe this should be a code comment. This is tribal knowledge that needs documenting somewhere.

Comment (mmlb) [reply ipxe/dhcp_options.go:71]: yep will be ther

Comment (mmlb) [reply ipxe/dhcp_options.go:71]: also in the commit message ;)

Comment (mmlb) [reply ipxe/dhcp_options.go:71]: done, see: [`e396a1d` (#149)](https://github.com/tinkerbell/boots/pull/149/commits/e396a1d3cafe8b8214389241df7abaed84d98295)

Comment (mmlb) [reply ipxe/dhcp_options.go:71]: woops well I got confused.

Comment (mmlb) [reply ipxe/dhcp_options.go:71]: sigh this was well documented in the commit that introduced this change but was wiped away when open sourcing boots and history was reset... :(. I'll fix that now.

Comment (mmlb) [reply ipxe/dhcp_options.go:72]: ok this too has been commented on in code and expanded on in the commit message.

File: dhcp/pxe.go [hunk 6]
```diff
@@ -94,6 +94,13 @@ func SetupPXE(ctx context.Context, rep, req *dhcp4.Packet) bool {
 		dhcplog.With("mac", req.GetCHAddr(), "xid", req.GetXID()).Info("no client GUID provided")
 	}
 
+	// Wanted by PXE clients to learn of PXE servers, not required but responses with PXEClient ClassID are preferred.
```
jacobweinstock, line 97: in `Figure 2-2 PXE Client Timeouts` and `Figure 2-3 PXE Client Response to DHCP Server Containing a Proxy DHCP Service` and a few other places in the [pxespec](http://www.pix.net/software/pxeboot/archive/pxespec.pdf) seem to indicate that a DHCP offer must have `PXEClient` in option 60. that would make this comment misleading.

ipxe/dhcp_options.go
jacobweinstock: thanks for the update. Would it be possible to add a direct pointer to where in Boots we are overriding this version? This is helpful but still a bit opaque.

dhcp/pxe.go
mmlb: I added more info in the commit message which has an excerpt from the spec:

Section 2.2.2 of the Intel PXE Spec states:

> The purpose of the timeouts is to ensure the PXE client gives precedence to servers supplying
> “PXEClient” specific configuration tags. The PXE boot ROM must function as a normal DHCP boot
> ROM in the absence of a PXE specific response. However, the PXE boot ROM must wait for
> specified times to see if a PXE response is available before using a non-PXE configuration.

Note _precedence_ in the spec, this lines up with our real world experience with *all* our nics in production. We we *not* setting the "PXEClient" class id when the first dhcp message comes in where we respond with filename:$our-iPXE-build.

ipxe/dhcp_options.go
mmlb: Yep, good call will add to comment.

dhcp/pxe.go
jacobweinstock: I totally understand that the real-world experience shows this option might not be needed but I believe the pxe spec shows that the server must send this option and I think we should implement the pxe spec regardless. All I'm suggesting is removing "not required" in the comment.

ipxe/dhcp_options.go
mmlb: Updated the commit, for ref its here: https://github.com/tinkerbell/boots/blob/main/ipxe/ipxe/build.sh#L43

dhcp/pxe.go
mmlb: The spec does not *require* it though. Section 2.2.2 Timeout calls out:

> However, the PXE boot ROM must wait for specified times to see if a PXE response is available before using a non-PXE configuration

This is a SHOULD, not MUST.

I'll reword the comment though to drop a mention of required or not though.

jacobweinstock: Thanks. 

Section 2.2.2 Timeout, is talking about the client waiting for PXE and non-PXE configurations. without PXEClient we would be in the class of non-PXE configuration. I don't think we want to implement a non-PXE configuration. especially since the pxe spec is about implementing a PXE configuration.

mmlb: I've updated the comment.

jacobweinstock: thanks, DHCP server/client interactions are terribly undocumented. I did happen to find [this](https://www.ibm.com/docs/en/tpmfi/7.1.1.16?topic=configuration-configuring-dhcp-server). Talks a little about the logic for option 60 with `PXEClient`.

mmlb: Its also from this time where the spec is written but failing smart/helpful was seen as good. Which leads to the spec and implementations to differ a lot :/ .

jacobweinstock (approved):

ipxe/dhcp_options.go
```diff
@@ -68,16 +68,18 @@ func Setup(rep *dhcp4.Packet) {
...
+// see VERSION_PATCH=255 in ./ipxe/ipxe/build.sh:43
+var ouriPXEVersion = []byte{1, 0, 255}
 
-// IsPacketIPXE returns bool depending on if the request originated with packet's ipxe build
-func IsPacketIPXE(req *dhcp4.Packet) bool {
+// IsOuriPXE returns bool depending on if the request originated from our custom built/embedded iPXE
+func IsOuriPXE(req *dhcp4.Packet) bool {
 	// TODO: make this actually check for iPXE and use ipxe' build system's ability to set name.
-	// This way we could set to something like "Packet iPXE" and then just look for that in the identifier sent in dhcp.
+	// This way we could set to something like "Boots iPXE" and then just look for that in the identifier sent in dhcp.
```
rgl, line 78: Can https://github.com/tinkerbell/boots/blob/4615ecf55f49f1666333a84b6009a0411bf4e2da/job/ipxe.go#L73 be modified to `Boots iPXE` too?

mmlb: yes good catch

PR #150, opened 2021-04-23
CI: Run tests and builds in two parallel jobs

mmlb:
## Description

Split ci.yaml so build and tests are two different jobs to get some better parallelism.

## Why is this needed

Hopefully makes builds a little faster (iPXE build still takes up most of the time though :( ).

## How Has This Been Tested?

CI

NA

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,4 @@
-name: For each commit and PR
+name: Test and Build Images Each Commit and PR
 on:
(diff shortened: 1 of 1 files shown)


markyjackson-taulia (dismissed):

mmlb:
Closing as now that we have self hosted runners on baremetal this is actually slower than before.

PR #158, opened 2021-05-04
Re-enable HTTP sanboot

### LINKED ISSUE #157: Re-enable http sanboot
When the iPXE build/config was reworked I got rid of sanboot. Turns out that breaks some EM usage and should be re-enabled.

## Expected Behaviour

`sanboot http://boot.ipxe.org/freedos/fdfullcd.iso` should work.

## Current Behaviour

sanboot command doesn't exist.

nshalman:
## Description

Explicitly enable HTTP sanboot per #157

## Why is this needed

It seems that the binaries stored in git-lfs likely were built with sanboot enabled and apparently some people seem to have been depending on that functionality.

Fixes: #157

## How Has This Been Tested?
I manually booted qemu, had it chainload to `undionly.kpxe`, and verified that I could drop to the ipxe shell and run
```
dhcp
sanboot http://boot.ipxe.org/freedos/fdfullcd.iso
``` 

Fixes #157

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/ipxe/.gitignore
+++ b/ipxe/.gitignore
@@ -1,1 +1,4 @@
 bindata.go
+bin/ipxe.efi
+bin/snp-nolacp.efi
+bin/undionly.kpxe
--- a/ipxe/bin/.gitignore
+++ b/ipxe/bin/.gitignore
@@ -1,3 +1,0 @@
-ipxe.efi
-snp-nolacp.efi
-undionly.kpxe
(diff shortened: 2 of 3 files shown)


nshalman:
Minimal test harness code in https://github.com/nshalman/boots/commit/b5aef63f56dd8e0216b107dd7d106c52ca10cbd0 
I can add that to this PR if desired.

mmlb (approved):

PR #164, opened 2021-06-11
Build IPXE on darwin easier:

jacobweinstock:
## Description

Currently building IPXE on a Mac is difficult and almost entirely manual, if not impossible even with the nix shell.nix.
This will leverage docker to make building IPXE on Mac just as easy as on Linux with the `make bindata` target call.

## Why is this needed

Building iPXE on Mac seems to not be a straightforward process.

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ all: help
 
 boots: cmd/boots/boots ## Compile boots for host OS and Architecture
 
-bindata: ipxe/bindata.go ## Build and generate embedded iPXE binaries
+bindata: bindata_by_os ## Build and generate embedded iPXE binaries
 
 crosscompile: $(crossbinaries) ## Compile boots for all architectures
 	
--- a/docs/DESIGN.md
+++ b/docs/DESIGN.md
@@ -104,7 +104,6 @@ The IPXE binaries are built from source and then embedded into the Boots Go bina
 
 ### Building the IPXE binary
 
-The scripts in this repo to build the IPXE binaries only work on Linux.
 The IPXE binaries from [https://github.com/ipxe/ipxe](https://github.com/ipxe/ipxe) are built via a Make target.
 
 ```make
--- a/rules.mk
+++ b/rules.mk
@@ -84,3 +84,14 @@ ipxe/ipxe/build/bin-arm64-efi/snp.efi ipxe/ipxe/build/bin-x86_64-efi/ipxe.efi ip
 	tar -xzf $< -C $(@D)
 	cp ${ipxeconfigs} $(@D)
 	cd $(@D) && ../../build.sh $$t ${ipxev}
+
+OSFLAG:= $(shell go env GOHOSTOS)
+# Build and generate embedded iPXE binaries based on OS
+bindata_by_os:
+ifeq (${OSFLAG},darwin)
+	rm -rf bin/go-bindata bin/goimports
+	docker run -it --rm -v ${PWD}:/code -w /code nixos/nix nix-shell --command "make ipxe/bindata.go"
+	rm -rf bin/go-bindata bin/goimports
+else
+	@$(MAKE) ipxe/bindata.go
+endif
--- a/shell.nix
+++ b/shell.nix
@@ -11,13 +11,16 @@ with pkgs;
 
 mkShell {
   buildInputs = [
+    curl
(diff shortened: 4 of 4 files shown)


markyjackson-taulia (approved):

Makefile
```diff
@@ -4,7 +4,15 @@ all: help
 
 boots: cmd/boots/boots ## Compile boots for host OS and Architecture
 
-bindata: ipxe/bindata.go ## Build and generate embedded iPXE binaries
```
nshalman, line 7: Why make the change here rather than in rules.mk?

nshalman (commented):
Have you tested this on Linux too?
Is this the best place to make the change?

jacobweinstock:
> Have you tested this on Linux too?
> Is this the best place to make the change?

Yes, tested locally in Linux. Worked. Is there a more preferable location?

Makefile
jacobweinstock: oh, maybe you mean a completely new target in rules.mk?

nshalman: This feels like a hack rather than a proper solution, but maybe that's what we need right now.
I'm not going to block this change, but I also feel uncomfortable explicitly approving it.

jacobweinstock: Would you mind helping me understand why you feel this is a hack?

docs/DESIGN.md
```diff
@@ -104,7 +104,6 @@ The IPXE binaries are built from source and then embedded into the Boots Go bina
 
 ### Building the IPXE binary
 
-The scripts in this repo to build the IPXE binaries only work on Linux.
```
mmlb, line 107: Is this accurate? I think the mac issue is really that at this nixpkgs pin the gcc cross compiler is not pre-built, otherwise it should work for mac (or would be easy to make work for mac). Would be good to see if its prebuilt in hydra in a different revision and/or ask in matrix about enabling for mac (if not pre-built on purpose).

mmlb (changes requested):

Makefile
jacobweinstock: sure I will move this to rules.mk. 

In my opinion, I don't believe having the Makefile only contain simple target definitions reduces or removes any complexity in our overall Make structure. It only moves the complexity and adds more code.

docs/DESIGN.md
jacobweinstock: Regardless of the gcc cross compiler package concerns, I am not able to build any IPXE binaries on my Mac. I've tried unsuccessfully to get it working. Below is just the latest error I'm running into.
```bash
nix-shell
make bindata
.
.
.
  [BUILD] bin-x86_64-efi/__divdi3.o
In file included from <command-line>:
./include/compiler.h:219: error: "__weak" redefined [-Werror]
  219 | #define __weak  __attribute__ (( weak, noinline ))
      | 
<built-in>: note: this is the location of the previous definition
cc1: all warnings being treated as errors
make[1]: *** [Makefile.housekeeping:943: bin-x86_64-efi/__divdi3.o] Error 1
make[1]: Leaving directory '/tinkerbell/boots/ipxe/ipxe/build/bin-x86_64-efi/ipxe-a5fb41873dc1dcce5dcc817bf53a0649c01f8cac/src'
make: *** [rules.mk:81: ipxe/ipxe/build/bin-x86_64-efi/ipxe.efi] Error 2
```

mmlb: @jacobweinstock want to try bumping ipxe to the latest git rev?

jacobweinstock: @mmlb sure. how do I do that?

mmlb: Update the values in https://github.com/tinkerbell/boots/blob/master/ipxev.mk, commit and the sha512sum of the archive tarball you can fetch from github.

jacobweinstock: @mmlb same error

mmlb: ok cool, we *could* also try a new nixpkgs (in case it has newer gcc that fixes this too)

jacobweinstock: @mmlb , tried that too. same error.

mmlb (approved):

PR #165, opened 2021-06-15
EM specific: VMware installer should respect boot_drive_hint for --firstdisk

nshalman:
## Description

When installing VMware, it can be helpful to provide a detailed hint as to which disk the VMware installer should use.
This commit adds support for `instance.boot_drive_hint` in the hardware metadata which will be read when generating the VMware kickstart file.

## Why is this needed

On some of our hardware types (our storage nodes in particular) the VMware installer will sometimes choose the wrong disk. Providing a better hint to the installer should prevent those issues.

Fixes: Internal ticket: PROV-16

## How Has This Been Tested?

- [x] Updated unit tests show that this works as expected
- [x] Manual retrieval of the kickstart file when running in a sandbox both with and without the `boot_drive_hint` also works as expected.

This currently only affects hardware whose plan slug starts with "s" which corresponds to EM storage hardware nodes.

To change that to affect all plans
```shell
git grep -l -E '& strings.HasPrefix(.*, "s")' | xargs sed -e 's|&& strings.HasPrefix(.*, "s") ||' -i
```
and verify that tests still pass and clean up any errors (e.g. unused import of "strings")

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/installers/vmware/kickstart-esxi.go
+++ b/installers/vmware/kickstart-esxi.go
@@ -3,6 +3,7 @@ package vmware
 import (
 	"io"
 	"net/http"
+	"strings"
 	"text/template"
 
 	"github.com/pkg/errors"
@@ -384,6 +385,10 @@ func rootpw(j job.Job) string {
 // We do not support anything other than ESXi 6.5 and above (os slug "vmware_esxi_6_5", "vmware_esxi_6_7", "vmware_esxi_7_0" etc)
 // full list of drive settings is listed https://packet.atlassian.net/browse/SWE-2385
 func determineDisk(j job.Job) string {
+	// currently limited to storage plans remove '&&.*"s")' to apply across all plans
+	if j.BootDriveHint() != "" && strings.HasPrefix(j.PlanSlug(), "s") {
+		return "--firstdisk=" + j.BootDriveHint()
+	}
 	switch j.PlanSlug() {
 	case "c1.small.x86",
 		"s1.large.x86",
--- a/packet/models.go
+++ b/packet/models.go
@@ -129,6 +129,8 @@ type Instance struct {
 	SSHKeys []string `json:"ssh_keys,omitempty"`
 	// CustomData
 	NetworkReady bool `json:"network_ready,omitempty"`
+	// BootDriveHint defines what the VMware installer should pass as the argument to "--firstdisk=".
+	BootDriveHint string `json:"boot_drive_hint,omitempty"`
 }
 
 // Device Full device result from /devices endpoint
--- a/installers/vmware/kickstart_script_test.go
+++ b/installers/vmware/kickstart_script_test.go
@@ -27,6 +27,7 @@ func TestDetermineDisk(t *testing.T) {
 func TestScriptKickstart(t *testing.T) {
 	manufacturers := []string{"supermicro", "dell"}
 	versions := []string{"vmware_esxi_6_0", "vmware_esxi_6_5", "vmware_esxi_6_7", "vmware_esxi_7_0"}
+	drive_hints := []string{"", "ST3120814A,mptsas,local"}
 	assert := require.New(t)
 	conf.PublicIPv4 = net.ParseIP("127.0.0.1")
 	conf.PublicFQDN = "boots-test.example.com"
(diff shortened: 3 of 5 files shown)


installers/vmware/kickstart-esxi.go
```diff
@@ -383,6 +383,9 @@ func rootpw(j job.Job) string {
 // We do not support anything other than ESXi 6.5 and above (os slug "vmware_esxi_6_5", "vmware_esxi_6_7", "vmware_esxi_7_0" etc)
 // full list of drive settings is listed https://packet.atlassian.net/browse/SWE-2385
 func determineDisk(j job.Job) string {
+	if j.BootDriveHint() != "" && j.PlanSlug()[:1] == "s" {
```
jacobweinstock, line 386: if `j.PlanSlug()` == "" then `j.PlanSlug()[:1]` will cause a panic. another option could be to use `strings.HasPrefix` which wont panic if `j.PlanSlug()` == "".

packet/models.go
```diff
@@ -129,6 +129,8 @@ type Instance struct {
 	SSHKeys []string `json:"ssh_keys,omitempty"`
 	// CustomData
 	NetworkReady bool `json:"network_ready,omitempty"`
+	// Boot Drive hint for VMware Installer
```
jacobweinstock, line 132: Typically a doc comment like this "[begins with the name of the item it describes](https://golang.org/doc/effective_go#commentary)".  Something like:
`// BootDriveHint defines what the VMware installer should use as the booting drive.`

installers/vmware/kickstart-esxi.go
nshalman: In this test hardness that should never happen, but this is still very good feedback and a change worth incorporating.

packet/models.go
nshalman: None of the doc comments in the preceding 20 lines seem to conform to that standard, but that's no reason not to start now. Will fix.

nshalman (commented):
Thanks for the review. I will address the feedback before this should be merged.

jacobweinstock (dismissed):
👍🏼

installers/vmware/kickstart-esxi.go
```diff
@@ -383,6 +384,9 @@ func rootpw(j job.Job) string {
 // We do not support anything other than ESXi 6.5 and above (os slug "vmware_esxi_6_5", "vmware_esxi_6_7", "vmware_esxi_7_0" etc)
 // full list of drive settings is listed https://packet.atlassian.net/browse/SWE-2385
 func determineDisk(j job.Job) string {
+	if j.BootDriveHint() != "" && strings.HasPrefix(j.PlanSlug(), "s") {
```
mmlb, line 387: Why do we want to limit it to just the `s` plans? If there's a boot drive hint we should honor it, if it shouldn't be there then we should update metadata to not have it at all.

nshalman: To limit blast radius at initial roll out. This is also the plan that is most in need of this feature.

mmlb: how about a comment then, this way if you forget to come back this wont stay around forever :D

nshalman: Will do. There are instructions in the PR description, but I'll add a short note just above that line in the code.

mmlb:
also needs a rebase

installers/vmware/kickstart-esxi.go
nshalman: Better?

jacobweinstock (dismissed):

installers/vmware/kickstart-esxi.go
mmlb: yep, thanks!

nicolerenee (dismissed):
LGTM :shipit:

jacobweinstock (dismissed):

jacobweinstock:
> # Testing Method
> 
> Set up a [bare metal sandbox](https://docs.tinkerbell.org/setup/equinix-metal-terraform/) 
> Boot the worker into OSIE, log in, and obtain the boot drive disk model with `lsblk -o model`
> In my example it is `Micron_5300_MTFD`
> 
> Important URLs:
> Primary PXE URL: `http://192.168.1.1:80/auto.ipxe`
> VMware Kickstart URL: `http://192.168.1.1:80/vmware/ks-esxi.cfg`
> 
> Make a copy of the worker's `hardware-data-0.json` to a new name and modify it to trigger the vmware installer:
> 
> For now because this feature is currently limited to storage hardware types, change `metadata.facility.plan_slug` to something that starts with `s`. For simplicity, I recommend `storage`
> 
> Minimal hardware data json chunk to be added to hardware to make it use vmware installer
> ```json
> {
>    "instance" : {
>       "boot_drive_hint" : "Micron_5300_MTFD",
>       "operating_system" : {
>          "distro" : "vmware",
>          "image_tag" : null,
>          "license_activation" : {
>             "state" : "unlicensed"
>          },
>          "name" : "VMware ESXi 7.0",
>          "slug" : "vmware_esxi_7_0",
>          "version" : "7.0"
>       }
>    }
> }
> ```
> 
> Here's what the diff looks like for me:
> ```diff before after
> 6c6
> <       "plan_slug": "c3.small.x86",
> ---
> >       "plan_slug": "storage",
> 9c9,21
> <     "instance": {},
> ---
> >     "instance": {
> >       "boot_drive_hint": "Micron_5300_MTFD",
> >       "operating_system": {
> >         "distro": "vmware",
> >         "image_tag": null,
> >         "license_activation": {
> >           "state": "unlicensed"
> >         },
> >         "name": "VMware ESXi 7.0",
> >         "slug": "vmware_esxi_7_0",
> >         "version": "7.0"
> >       }
> >     },
> ```
> 
> You can now `tink hardware push < modified.json`
> 
> From the worker do curls to confirm:
> ```
> curl -s http://192.168.1.1/auto.ipxe | grep -i vmware
> curl -s http://192.168.1.1:80/vmware/ks-esxi.cfg | grep -i firstdisk
> ```
> 
> Before:
> ```
> localhost:~# curl -s http://192.168.1.1/auto.ipxe | grep -i vmware
> set base-url http://install.packet.net/vmware/esxi-7.0.0
> kernel ${base-url}/mboot.c32 -c ${base-url}/boot.cfg ks=${tinkerbell}/vmware/ks-esxi.cfg netdevice=0c:42:a1:8f:97:1a ksdevice=0c:42:a1:8f:97:1a
> localhost:~# curl -s http://192.168.1.1:80/vmware/ks-esxi.cfg | grep -i firstdisk
> install --firstdisk --overwritevmfs
> ```
> 
> Now upgrade boots to the version in this PR and modify `docker-compose.yml` so that the `MIRROR_HOST` line looks like:
> ```
> MIRROR_HOST: install.packet.net
> ```
> 
> To upgrade boots, on your development machine you can run:
> ```
> make image
> docker image save boots:latest | ssh root@${PROVISIONER:?} docker image load
> ```
> 
> And on the provisioner:
> ```
> cd tink/deploy
> . ../.env
> export TINKERBELL_TINK_BOOTS_IMAGE=boots:latest
> docker-compose up -d
> ```
> 
> Now from the worker, do the curls again:
> ```
> localhost:~# curl -s http://192.168.1.1/auto.ipxe | grep -i vmware
> set base-url http://install.packet.net/vmware/esxi-7.0.0
> kernel ${base-url}/mboot.c32 -c ${base-url}/boot.cfg ks=${tinkerbell}/vmware/ks-esxi.cfg netdevice=0c:42:a1:8f:97:1a ksdevice=0c:42:a1:8f:97:1a
> localhost:~# curl -s http://192.168.1.1:80/vmware/ks-esxi.cfg | grep -i firstdisk
> install --firstdisk=Micron_5300_MTFD --overwritevmfs
> ```
> 
> After a reboot of the worker, it *should* try to install vmware, but I couldn't see anything interesting on the SOS console during the vmware install...

This is a great doc! Might be something worth putting in the repo.

tstromberg (approved):

PR #168, opened 2021-06-24
vmware: Add a pause in pre-install to quell network devices

truongmd:
## Description

Adding a sleep to allow network devices to settle

## Why is this needed

We see a lot of install failures that look to be network interface not settling fast enough

Fixes: #

## How Has This Been Tested?
Custom iPXE test deploy

30 sec longer install time

## Checklist:

I have:

- [x] added unit or e2e tests
- [x] provided instructions on how to upgrade


mmlb:
I'm just gonna close this and do it on my own branch as I can't push to this for some reason...

PR #169, opened 2021-06-24
vmware: Add a pause in pre-install to quell network devices

mmlb:
## Description

Adding a sleep to allow network devices to settle

## Why is this needed

We see a lot of install failures that look to be network interface not settling fast enough

## How Has This Been Tested?
Custom iPXE test deploy

## How are existing users impacted? What migration steps/scripts do we need?

30 sec longer install time

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required)
- [x] added unit or e2e tests
- [ ] provided instructions on how to upgrade

PR #170, opened 2021-06-24
Add 30 second sleep to ESXi pre-install to quell network devices

truongmd:
Signed-off-by: My D. Truong <truongmd@gmail.com>

## Description

Adding a sleep to allow network devices to settle

## Why is this needed

We see a lot of install failures that look to be network interface not settling fast enough

Fixes: #

## How Has This Been Tested?

Custom iPXE test deploy

30 sec longer install time

## Checklist:

I have:

- [x] added unit or e2e tests
- [x] provided instructions on how to upgrade


mmlb:
meh #169 is almost done building, going with that one instead

PR #159, opened 2021-05-05
Automated test of ipxe config to confirm sanboot functionality

nshalman:
## Description

While this doesn't directly exercise the ipxe binaries we bake into boots,
it does use the ipxe configs to build an `ipxe.lkrn` binary to use for testing features.
This test uses expect and qemu to exercise the sanboot feature.

## Why is this needed

Related to #157

## How Has This Been Tested?

`make test-sanboot` succeeds.
Additionally, cherry-picking this commit onto 5d563d8 fails as expected.

## How are existing users impacted? What migration steps/scripts do we need?

N/A

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required)
- [x] added unit or e2e tests
- [ ] provided instructions on how to upgrade

PR #126, opened 2021-02-01
Reproducible iPXE builds

mmlb:
## Description

Build iPXE so that the binaries are reproducible (verified so on the same machine)

## Why is this needed

Helps with debugging issues.

## How Has This Been Tested?

Compiles.

PR #171, opened 2021-06-29
Add Docker Buildx back into the ci workflow

nshalman:
## Description

Fix the docker-buildx ci workflow. This is a revert of one of the changes in #148.

## Why is this needed

#165 fails to build without this change
#159's build is also failing, likely for the same reason. 
Also #126 is probably blocked for the same reason...

## How Has This Been Tested?

Applied this commit to #165 and the builds then succeeded.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,6 +20,8 @@ jobs:
         uses: cachix/install-nix-action@018abf956a0a15673dae4932ae26f0f071ac0944
         with:
           nix_path: nixpkgs=channel:nixpkgs-unstable
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
       - name: Non Go formatters and linters
         run: ./.github/workflows/ci-non-go.sh
       - name: Install Go


.github/workflows/ci.yaml
```diff
@@ -20,6 +20,10 @@ jobs:
         uses: cachix/install-nix-action@018abf956a0a15673dae4932ae26f0f071ac0944
         with:
           nix_path: nixpkgs=channel:nixpkgs-unstable
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          version: "v0.4.1"
```
matoszz, line 26: unless there's some specific version dep here, use the generic version

matoszz (changes requested):
If the desire is for faster builds overall (hence the removal of buildx) I would also add the cache action `actions/cache@v2` - you can control this based on tag if you'd like, and use it throughout the pipeline, e.g.

```
      - name: Set up Docker Buildx
        id: buildx
        uses: docker/setup-buildx-action@v1
      - name: Get tag
        uses: tenhaus/get-release-or-tag@v2
        id: tag
      - name: Docker meta
        id: docker_meta
        uses: crazy-max/ghaction-docker-meta@v1
        with:
          images: quay.io/tinkerbell/boots
          tag-custom: ${{ steps.tag.outputs.tag }}
      - name: Cache Docker layers
        uses: actions/cache@v2
        with:
          path: /tmp/.buildx-cache
          key: ${{ runner.os }}-boots-cache
          restore-keys: |
            ${{ runner.os }}-boots-cache

```
In your build and push docker image, you can then use `cache-from: type=local,src=/tmp/.buildx-boots-cache`

.github/workflows/ci.yaml
nshalman: Fixed.

nshalman:
> If the desire is for faster builds overall (hence the removal of buildx) I would also add the cache action `actions/cache@v2` - you can control this based on tag if you'd like, and use it throughout the pipeline, e.g.

The boots docker container is so minimalist that I don't think this is a particular concern.

markyjackson-taulia (approved):

matoszz (approved):

PR #173, opened 2021-07-01
log discovery requests

mikemrm:
Discovery requests call an external service, by logging this, it's
easier to track a request when it takes this path.

### DIFF
--- a/cmd/boots/main.go
+++ b/cmd/boots/main.go
@@ -74,7 +74,7 @@ func main() {
 		mainlog.With("envvar", "API_AUTH_TOKEN").Fatal(err)
 		panic(err)
 	}
-	client, err = packet.NewClient(consumer, auth, apiBaseURL)
+	client, err = packet.NewClient(l, consumer, auth, apiBaseURL)
 	if err != nil {
 		mainlog.Fatal(err)
 	}
--- a/packet/client.go
+++ b/packet/client.go
@@ -13,6 +13,7 @@ import (
 
 	cacherClient "github.com/packethost/cacher/client"
 	"github.com/packethost/pkg/env"
+	"github.com/packethost/pkg/log"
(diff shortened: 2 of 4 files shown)


cmd/boots/main.go
```diff
@@ -73,6 +73,9 @@ func main() {
 	if err != nil {
 		mainlog.Fatal(err)
 	}
+
+	client.Logger = l
```
mmlb, line 77: would prefer to have NewClient gain a log.Logger param

mmlb (changes requested):

packet/client.go
```diff
@@ -28,9 +29,10 @@ type Client struct {
 	authToken      string
 	hardwareClient hardwareGetter
 	workflowClient tw.WorkflowServiceClient
+	Logger         log.Logger
```
mmlb, line 32: make private

packet/client.go
```diff
@@ -28,9 +29,10 @@ type Client struct {
...
 }
 
-func NewClient(consumerToken, authToken string, baseURL *url.URL) (*Client, error) {
+func NewClient(consumerToken, authToken string, baseURL *url.URL, logger log.Logger) (*Client, error) {
```
mmlb, line 35: lets do it as first param.

mmlb (changes requested):

packet/client.go
mikemrm: gah, you're right!

tstromberg (approved):

mmlb (dismissed):
just needs a rebase @mikemrm

mmlb:
> just needs a rebase @mikemrm

nm I was able to do the rebase and push

jacobweinstock:
Hey @mikemrm, looks like there's just a single file merge conflict but other than that should be could to go.

PR #177, opened 2021-07-21
Add standalone mode

tobert:
## Description

This PR adds a "standalone" mode to boots so it can operate without a Packet API or Tinkerbell server. Instead of contacting a backend service, it loads a JSON file at startup and uses only information from there. This might not be terribly useful in the datacenter but makes local iteration on various features much easier and quicker since boots can run without any dependencies.

## Why is this needed

This makes local iteration much faster. It also could be useful in labs where the full power of Tinkerbell is more than what the user wants.

## How Has This Been Tested?

Existing tests pass. I validated it all in a pair of docker containers, one running boots and the other running `dhclient -4 -d -w`

Shouldn't impact anyone not using the "standalone" path.

## Checklist:

I have:

- [ ] added unit or e2e tests


jacobweinstock (commented):
This is really great! thank you for the work on this. I think it's a very useful addition. Wanted to discuss the `panic`s in `models_standalone.go`. any chance those panics could be moved up higher in the stack? ideally, I'd love to see the [main func](https://github.com/tinkerbell/boots/blob/cb0290f8d14e1e168c32ce425d0c616a08742d7f/cmd/boots/main.go#L41) be the only thing that panics. What are your thoughts?

tobert:
> Wanted to discuss the `panic`s in `models_standalone.go`. any chance those panics could be moved up higher in the stack? What do you think about having the [main func](https://github.com/tinkerbell/boots/blob/cb0290f8d14e1e168c32ce425d0c616a08742d7f/cmd/boots/main.go#L41) be the only thing that panics?

Fair enough. Most of those methods do not have a way to return error so I chose panic to make it clearer that the limitation is because the code is incomplete. If I or someone else rolls through and adds multi-interface support to that code most or all of them would go away. My goal was to get as quickly as possible to demonstrate the idea works, and find out if y'all want it. Assuming you do, I can lay in the multiple interface support and eliminate most of them.

tobert:
@jacobweinstock the panics are gone now :)

test/standalone-hardware.json
```diff
@@ -0,0 +1,88 @@
...
+        ],
+        "userdata": "lol hi"
+      },
+      "custom": {
+        "private_subnets": [
+          "192.168.192.0/24"
+        ]
+      },
+      "facility": null,
+      "provisioner_engine": "standalone"
```
jacobweinstock, line 85: when using this example file and the updated README.md, I was unable to get Boots to serve DHCP. When a DHCP request comes in Boots checks this value against this environment variable: `PROVISIONER_ENGINE_NAME`. To get it working i needed to either change the `provisioner_engine` value to `""` or set the environment variable `PROVISIONER_ENGINE_NAME` to `standalone` when running Boots.

tobert: Yeah, that key doesn't do anything, only the envvar is respected. It appears in the JSON because I crafted it from the existing structs in the codebase, rather than designing my own structs, so there are some dead fields. I'll remove some of the cruft from the config so it's less confusing.

jacobweinstock (approved):

tobert:
One of these days I will remember to sign-off on commit 😅

jacobweinstock (approved):

PR #188, opened 2021-08-05
Add design philosophy doc:

jacobweinstock:
## Description

This makes it clear how to approach code design and gives
us a document to point to for discussions and/or concerns
that may arise.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


jacobweinstock:
> Sounds good. What would you think about proposing this to https://docs.tinkerbell.org ?

PR into https://docs.tinkerbell.org instead of here or in addition to here?

tstromberg:
Either is fine - just an idea.

Issue #190, opened 2021-08-05
Proposed Boots roadmap


jacobweinstock:
The following is a proposed roadmap of work items for Boots.
They are not ordered in terms of priority.
I don't know if we can get a Github project created in this repo to maybe track these.
If so, I can create all the tickets.

- [x] define design philosophy
- [x] turn up golangci-lint
- [ ] document coding style guide
- [x] document data model relationship to code logic
- [x] installer registration explicit
- [x] all environment vars defined in func main
- [x] use github.com/peterbourgon/ff for flags, env vars, config file
- [x] reorg package structure
- [x] proper context passing
- [x] document versioning strategy for the codebase
- [ ] incorporate goreleaser
- [ ] customizable DHCP options
- [x] customizable kernel args
- [x] move default git branch to main
- [x] move all panics to func main
- [x] move logging to github.com/go-logr/logr
- [x] move osie installer to hook
- [x] add DHCPProxy functionality
- [ ] add DHCP pools
- [ ] add hardware discovery functionality
- [x] add custom ipxe script with network retries and custom user-class name
- [x] add stack stand up capability
- [x] remove/avoid global/package level variables
- [x] remove cacher dependency
- [x] remove git-lfs dependency
- [x] remove installers other than a default
- [ ] remove syslog
- [x] remove hardware discovery endpoint
- [x] remove Equinix Metal specific code
  - [x]   dual provisioner lookup
  - [x]   hollow
  - [x]   EM api calls
  - [x]   EM specific http endpoints
  - [x]   etc
- [x] refactor to make all funcs/methods explicitly require their dependencies
- [ ] add functional testing
- [x] Increased unit testing
- [x] #210

displague:
@jacobweinstock I think we should create trackable issues for each of these items (although some items may fit into the same issue). This roadmap issue contains items that could extend for sometime. The bulleted items are missing context that would be provided in a dedicated issue.  Those issues can be linked to in the bulleted list above. I think we could then create a meaningful epic of this issue, targeting a specific goal achievable by completing some set of the bulleted items (roadmap is not a specific goal). As you point out, instead of using a tracking issue, we could make a project geared towards specific goals if we have a clear goal and a path to achieve it.

I don't have the option of creating a project, but I can create a milestone and I think this could work well enough for now. 

[...]

Issue #110, opened 2020-12-03
When an IPXe URL is a data-url, expand the url into ipxe content


displague:
## Expected Behaviour

When providing an IPXe URL, that is data-uri encoded, boots should expand this URI into the decoded content and pass this to IPXE.  This would be equivalent to `#!ipxe` userdata, but frees the userdata metadata to be consumed as userdata (cloud-init, ignition, etc).

## Current Behaviour

Data URIs are not handled.

## See also

* https://github.com/ipxe/ipxe/issues/174

jacobweinstock:
sorry for the dumb questions, i don't think i fully understand the value here.
Is this just a different format to provide the ipxe script?
what data would be encoded here? the kernel? 
how is this any different than a normal ipxe script?
would you mind expanding on this statement? "frees the userdata metadata to be consumed as userdata"

displague:
> would you mind expanding on this statement? "frees the userdata metadata to be consumed as userdata"

When providing an IPXE Script, you can currently take advantage of the userdata field with a value of `#!ipxe\n...`.  In doing so, you lose the ability to take advantage of userdata for the intended purpose, which is for tools like cloud-init and afterburn to have a predefined location to discover platform-specific metadata.  While this location is hardcoded into some systems (afterburn) it is generally configurable in the kernel command line.

We provide only one custom URL that can be consumed in this way.  When that facility is repurposed to provide an IPXE script, we lose ability to treat the userdata field as cloud-config, kickstart config, multi-part mime cloud-configs and scripts, etc.

displague:
> what data would be encoded here? the kernel?

The IPXE Script would be encoded.

To provide an ipxe_script_url today, you must have a web server that is accessible to the node, which may have networking restrictions that prevent that.  But allowing for public internet routing and assuming I have a web-host available and my data is not sensitive (too many assumptions), I can park my IPXE Script URL on Github.

`ipxe_script_url=https://gist.githubusercontent.com/displague/cd5938eb7584dd1c1cc76c28a3f9c762/raw/67e2047ca6304a8aaec80e49fdf38bf4da4f5a48/centos8-stream.ipxe`

[...]

PR #192, opened 2021-08-10
support a more generic ipxe installer

mikemrm:
## Description

Allows more generic support for ipxe installers for installation of an OS which cannot be modified to work with the OSIE installation method.

The new ipxe installer works similar to custom_ipxe however differs in that it is selected earlier in the process for different handling of data and which is used for any os which has the installer set to `ipxe`.

## Why is this needed

Some os installation processes have unique installation environments which do not lend themselves easily to osie liveos / workflow setup as seen with other installers in boots like vmware and coreos. However instead of hard coding a new OS into boots, this method allows for those dynamic configurations to be handled separately by allowing the installer to be set to ipxe which supports chaining or injecting a script to boot the installer.

The model has changed for `packet.OperatingSystem` to now have two additional optional fields `Installer` and `InstallerData`.

The ipxe installer is configurable through the `InstallerData` field which it expects to be a string encoded JSON blob of which two fields may be specified `chain` which should be a full URL to a valid iPXE script and `script` which allows for injecting an ipxe script directly.

Example of both configurable fields:

```json
{
  "chain": "https://boot.example.com/unique_os_installer.ipxe",
  "script": "echo Welcome to unique os installer\nshell\n"
}
```

Example using iPXE installer with chain loading:

```json
{
  "distro": "Unique OS",
  "os_slug": "unique_os",
  "slug": "unique_os_7",
  "version": "7",
  "installer": "ipxe",
  "installer_data": "{\"chain\": \"https://boot.example.com/unique_os_installer.ipxe\"}"
}
```

No changes to existing functionality, only adding additional support.


jacobweinstock:
Hey @mikemrm, I'm not seeing how this is different than the existing `custom_ipxe`? `custom_ipxe` can already handle ipxe scripts via a [URL to chain load](https://github.com/tinkerbell/boots/blob/94f43947757e6fae068f4265b89eeda1bf04282c/installers/custom_ipxe/main.go#L22) or just a [custom IPXE script](https://github.com/tinkerbell/boots/blob/94f43947757e6fae068f4265b89eeda1bf04282c/installers/custom_ipxe/main.go#L20).

mikemrm:
Correct @jacobweinstock however this differs in how it is triggered. The `custom_ipxe` is only used when an OperatingSystem distro is set to `custom_ipxe` which it then looks at a field attached to the instance `ipxe_script_url`.

The trouble is that this process is only triggered due to the distro being `custom_ipxe` so if we needed to support a new OS which would have a distro of `My OS` (`my_os`) which needed to chain load an ipxe script. We would have to overwrite the distro field sent to cacher/tink with `custom_ipxe` which is undesireable.

This change makes this path more direct by not relying faking / overriding a field.

mikemrm:
> Also, I've put out a [proposal ](https://github.com/tinkerbell/boots/issues/190) that includes deprecating all installers besides a default. I believe there has been an unofficial desire amongst the community to deprecate all installers too.

I agree with that desire and I believe this functionality would allow us to remove the other installers leaving only this method.

jacobweinstock:
@mikemrm , there are actually 2 fields that can trigger a Boots Installer. `metadata.instance.operating_system.slug` and `metadata.instance.operating_system.distro`. `slug` is [checked first](https://github.com/tinkerbell/boots/blob/94f43947757e6fae068f4265b89eeda1bf04282c/job/ipxe.go#L80) and then `distro` is [checked](https://github.com/tinkerbell/boots/blob/94f43947757e6fae068f4265b89eeda1bf04282c/job/ipxe.go#L85).

If I understand the use-case correctly, if the `slug` is set to be `custom_ipxe`, the `distro` can be anything (`my_os`) for example.

mikemrm:
Correct, it does appear custom_ipxe is actually only being registered for a slug. Meaning slug is the field that would need to be set.

This however doesn't change the issue, as slugs are typically understood as a unique string. Which in this case is the Operating System Version slug which an OS may have many OSVs.

mmlb (commented):
chaining and script can both be done with custom_ipxe, why do we need this instead of custom_ipxe?

mmlb:
> chaining and script can both be done with custom_ipxe, why do we need this instead of custom_ipxe?

well I should refresh before commenting it seems

jacobweinstock:
@mikemrm and I talked offline. (@mikemrm correct me if I misspeak here at all!) Here's the synopsis. There is a desire to add an additional hardware data field `metadata.instance.operatiing_system.Installer` that will be checked before `slug` and `distro`. Also, there is a desire to add another new add hardware data field: `metadata.instance.operatiing_system.IntallerData`. This field would be a `string` but expected to be a JSON string of `{"chain": "", "script":""}`.
The reasoning behind this is that it makes sense for the custom business logic internal to Equinix metal that interacts with hardware data via tink/cacher. As the hardware data model in Tink is arbitrary and not enforced, I don't see why this shouldn't be allowed. I did ask Mike to compromise and not create a new Boots Installer but to modify `custom_ipxe` without breaking backwards compatibility. Mike said he will see what he can do and update this PR.

mmlb:
Do we really need support for both `chain` and `script` fields? Chain can already by done by filling into script.

mikemrm:
> Do we really need support for both `chain` and `script` fields? Chain can already by done by filling into script.

Script was included to make it easier to push an exact script rather than relying on an externally hosted one. This is just a beneficial addition I can see useful for many things including testing without relying on external services.

installers/custom_ipxe/main.go
```diff
@@ -1,24 +1,96 @@
...
+		cfg, err = ipxeConfigFromJob(j)
+		if err != nil {
+			s.Echo("Failed to decode installer data")
+			s.Shell()
+			logger.Error(err, "decoding installer data")
+
+			return
+		}
+	} else {
+		cfg = &Config{}
```
jacobweinstock, line 37: don't think this is needed, is it?

installers/custom_ipxe/main.go
```diff
@@ -1,24 +1,96 @@
...
 
 func init() {
-	job.RegisterSlug("custom_ipxe", bootScript)
+	job.RegisterInstaller("ipxe", ipxeScript)
+	job.RegisterSlug("custom_ipxe", ipxeScript)
 }
 
-func bootScript(j job.Job, s *ipxe.Script) {
+func ipxeScript(j job.Job, s *ipxe.Script) {
+	logger := installers.Logger("ipxe")
```
jacobweinstock, line 19: this won't be accurate when called with the `custom_ipxe` slug. anything that can be done to accommodate both? maybe `job.RegisterInstaller("custom_ipxe", ipxeScript)`?

installers/custom_ipxe/main.go
```diff
@@ -1,24 +1,96 @@
...
+			cfg.Chain = j.IPXEScriptURL()
+		}
+	}
+
+	IpxeScriptFromConfig(logger, cfg, j, s)
+}
+
+func IpxeScriptFromConfig(logger log.Logger, cfg *Config, j job.Job, s *ipxe.Script) {
+	if err := cfg.validate(); err != nil {
+		s.Echo("Invalid ipxe configuration")
```
jacobweinstock, line 51: wouldn't using the `err.String()` be more accurate and informative here?

installers/custom_ipxe/main.go
```diff
@@ -1,24 +1,96 @@
...
+		s.Shell()
+		logger.Error(err, "validating ipxe config")
+
+		return
+	}
+
 	s.PhoneHome("provisioning.104.01")
 	s.Set("packet_facility", j.FacilityCode())
 	s.Set("packet_plan", j.PlanSlug())
```
jacobweinstock, line 61: should these be called at the beginning of `IpxeScriptFromConfig`?

packet/models.go
```diff
@@ -221,11 +221,13 @@ type IP struct {
...
-	Version  string `json:"version"`
-	ImageTag string `json:"image_tag"`
-	OsSlug   string `json:"os_slug"`
+	Slug          string `json:"slug"`
+	Distro        string `json:"distro"`
+	Version       string `json:"version"`
+	ImageTag      string `json:"image_tag"`
+	OsSlug        string `json:"os_slug"`
+	Installer     string `json:"installer,omitempty"`
+	InstallerData string `json:"installer_data,omitempty"`
```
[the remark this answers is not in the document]
mikemrm: Yeah, I can change this to a `map[string]interface{}` to keep it generic but enforce a base format. My hope is to keep it generic in case there is a future case for a different generic installer. Granted, that may be unnecessary. Thoughts?

installers/custom_ipxe/main.go
mikemrm: `cfg` at this point would be nil. But I can create it during the setting of the value instead of before.

installers/custom_ipxe/main.go
mikemrm: I can just change it to `custom_ipxe` since that's the package the logs will be coming from.

jacobweinstock:
I think it's important to note the following. This functionality can already be accomplished via the existing `custom_ipxe` Installer. In my opinion, what this PR does do is create a more clear mapping from hardware data to Boots installer.
New:
`metadata.instance.operating_system.installer` -> Boots Installer
Existing:
`metadata.instance.operating_system.slug` -> Boots Installer
`metadata.instance.operating_system.distro` -> Boots Installer

installers/custom_ipxe/main.go
mikemrm: I had purposefully made the ipxe output more generic and the logs more detailed, but I can just print the error here. Sure

installers/custom_ipxe/main.go
mikemrm: I don't want to phone home unless we properly handed off the ipxe as expected.

installers/custom_ipxe/main.go
```diff
@@ -1,24 +1,96 @@
...
 
-	if strings.HasPrefix(j.UserData(), "#!ipxe") {
-		s.AppendString(strings.TrimPrefix(j.UserData(), "#!ipxe"))
+	if cfg.Chain != "" {
+		s.Chain(cfg.Chain)
+	} else if cfg.Script != "" {
+		s.AppendString(strings.TrimPrefix(cfg.Script, "#!ipxe"))
```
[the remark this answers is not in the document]
mikemrm: Line 40 doesn't need TrimPrefix since it just sets `cfg.Script` which here we then TrimPrefix. Making this trim happen only once instead of scattered in multiple places.

installers/custom_ipxe/main.go
```diff
@@ -1,24 +1,96 @@
...
 	} else {
-		s.Chain(j.IPXEScriptURL())
+		s.Echo("Unknown ipxe config path")
```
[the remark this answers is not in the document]
mikemrm: Yeah, this was more for completeness, but I can remove it if preferred.

packet/models.go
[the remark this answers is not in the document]
mikemrm: Will go ahead and just use `map[string]string` if for some reason we need something else in the future we can handle it then.

installers/custom_ipxe/main.go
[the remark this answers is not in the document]
jacobweinstock: ah, yes I see now. you're right. thanks!

installers/custom_ipxe/main.go
jacobweinstock: ok. looks like line 65 is in a similar situation, we didn't properly hand off ipxe as expected? we are shelling out to ipxe there just like in line 49.

installers/custom_ipxe/main.go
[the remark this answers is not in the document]
jacobweinstock: completeness? sorry, not following.

packet/models.go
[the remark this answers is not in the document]
jacobweinstock: http://c2.com/xp/YouArentGonnaNeedIt.html
Any chance you would reconsider a proper struct?

installers/custom_ipxe/main.go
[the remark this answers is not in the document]
mikemrm: It was unnecessary. removed.

installers/custom_ipxe/main.go
```diff
@@ -3,22 +3,71 @@ package custom_ipxe
...
+			cfg = &packet.InstallerData{Script: j.UserData()}
+		} else {
+			cfg = &packet.InstallerData{Chain: j.IPXEScriptURL()}
+		}
+	}
+
+	IpxeScriptFromConfig(logger, cfg, j, s)
+}
+
+func IpxeScriptFromConfig(logger log.Logger, cfg *packet.InstallerData, j job.Job, s *ipxe.Script) {
```
jacobweinstock, line 47: should this function be unexported?

jacobweinstock (changes requested):
Thank you for being open to change so much!

installers/custom_ipxe/main.go
mikemrm: good point, this doesn't need to be anymore. Will fix

jacobweinstock (approved):
just a DCO issue to address and this should be good to go!

installers/custom_ipxe/main.go
```diff
@@ -3,22 +3,71 @@ package custom_ipxe
...
+	"github.com/packethost/pkg/log"
+	"github.com/tinkerbell/boots/installers"
 	"github.com/tinkerbell/boots/ipxe"
 	"github.com/tinkerbell/boots/job"
+	"github.com/tinkerbell/boots/packet"
 )
 
 func init() {
-	job.RegisterSlug("custom_ipxe", bootScript)
+	job.RegisterInstaller("ipxe", ipxeScript)
```
jacobweinstock, line 14: would `custom_ipxe` here make sense? this is the custom_ipxe package and installer.

mikemrm: sure i can make it custom_ipxe

mikemrm:
> > > Do we really need support for both `chain` and `script` fields? Chain can already by done by filling into script.
> > 
> > 
> > Script was included to make it easier to push an exact script rather than relying on an externally hosted one. This is just a beneficial addition I can see useful for many things including testing without relying on external services.
> 
> Right, so why don't we just support only `script` and if the use wants to chain they can provide the script that does so.

Besides the fact that it supports the two code paths of the original custom_ipxe method. I think it also makes an easy way to do the chain loading and adds minimal code. However if you feel this cannot be, I can remove.

installers/custom_ipxe/main.go
```diff
@@ -3,22 +3,71 @@ package custom_ipxe
...
+func ipxeScript(j job.Job, s *ipxe.Script) {
+	logger := installers.Logger("custom_ipxe")
+	if j.InstanceID() != "" {
+		logger = logger.With("instance.id", j.InstanceID())
+	}
+
+	var cfg *packet.InstallerData
+	var err error
+
+	if j.OperatingSystem().Installer == "ipxe" {
```
mmlb, line 27: this should be `custom_ipxe` right?

mmlb:
> > > > Do we really need support for both `chain` and `script` fields? Chain can already by done by filling into script.
> > > 
> > > 
> > > Script was included to make it easier to push an exact script rather than relying on an externally hosted one. This is just a beneficial addition I can see useful for many things including testing without relying on external services.
> > 
> > 
> > Right, so why don't we just support only `script` and if the use wants to chain they can provide the script that does so.
> 
> Besides the fact that it supports the two code paths of the original custom_ipxe method. I think it also makes an easy way to do the chain loading and adds minimal code. However if you feel this cannot be, I can remove.

Yeah I think this makes sense, it also makes explicit what was implicit before (either you feed in an `ipxe` script or we'll chainload what is presumably a url).

jacobweinstock (dismissed):

installers/custom_ipxe/main.go
```diff
@@ -3,22 +3,71 @@ package custom_ipxe
...
+	job.RegisterInstaller("custom_ipxe", ipxeScript)
+	job.RegisterSlug("custom_ipxe", ipxeScript)
 }
 
-func bootScript(j job.Job, s *ipxe.Script) {
+func ipxeScript(j job.Job, s *ipxe.Script) {
+	logger := installers.Logger("custom_ipxe")
+	if j.InstanceID() != "" {
+		logger = logger.With("instance.id", j.InstanceID())
+	}
```
mmlb, line 22: the job should have its own logger already with these fields (and more) already populated, so this should just be:
```go
logger := job.logger.With("installer", "custom_ipxe")
```

installers/custom_ipxe/main.go
```diff
@@ -3,22 +3,71 @@ package custom_ipxe
...
 func init() {
-	job.RegisterSlug("custom_ipxe", bootScript)
+	job.RegisterInstaller("custom_ipxe", ipxeScript)
+	job.RegisterSlug("custom_ipxe", ipxeScript)
 }
 
-func bootScript(j job.Job, s *ipxe.Script) {
+func ipxeScript(j job.Job, s *ipxe.Script) {
+	logger := installers.Logger("custom_ipxe")
+	if j.InstanceID() != "" {
```
mmlb, line 20: Hmm we should probably return a s.Shell if there's no instance id right?

installers/custom_ipxe/main.go
```diff
@@ -3,22 +3,71 @@ package custom_ipxe
...
+
+	var cfg *packet.InstallerData
+	var err error
+
+	if j.OperatingSystem().Installer == "custom_ipxe" {
+		cfg = j.OperatingSystem().InstallerData
+		if cfg == nil {
+			s.Echo("Installer data not provided")
+			s.Shell()
+			logger.Error(err, "empty installer data")
```
mmlb, line 32: err is currently nil here, actually this looks like the only use of `err` at all so no need for `var err error` right?

mmlb (changes requested):

installers/custom_ipxe/main.go
mikemrm: good catch, with all the changes this is no longer needed.

installers/custom_ipxe/main.go
mikemrm: I see your point, will update.

jacobweinstock (approved):

displague:
I'm late to the party, but I have brought up previously that a data uri should be supported as a way to offer ipxe scripts. #110 

This would be a minimal change that introduces added functionality within the existing spec.

Thoughts, @mikemrm?

mikemrm:
That is an interesting idea @displague however I'm not sure how common it is for someone to want to data uri encode a script and would be interested to see others thoughts about supporting it.

displague:
> however I'm not sure how common it is for someone to want to data uri encode a script

I imagine users are not writing templates by hand and would use CLIs and APIs to get the value in place.

We have a URL field and data URLs are an established pattern for providing any data as a URL. Rather than fetch the resource over HTTP, ftp, nntp, gopher, we have the data contained in the URL and the field can be said to support HTTP, HTTPS, and data URLs.

Existing APIs and CLIs that wrap the Tinkerbell API and pass the IPXE URL to Tinkerbell will get this functionality for free.

installers/custom_ipxe/main.go
```diff
@@ -3,22 +3,70 @@ package custom_ipxe
...
+			s.Echo("Installer data not provided")
+			s.Shell()
+			logger.Error(ErrEmptyIpxeConfig, "installer data not provided")
+
+			return
+		}
+	} else if strings.HasPrefix(j.UserData(), "#!ipxe") {
+		cfg = &packet.InstallerData{Script: j.UserData()}
+	} else if j.IPXEScriptURL() != "" {
+		cfg = &packet.InstallerData{Chain: j.IPXEScriptURL()}
```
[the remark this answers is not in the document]
mmlb: @displague lets give it a go in a different PR.

mmlb (approved):

PR #193, opened 2021-08-10
more tracing features

tobert:
## Description

Adds (more) tracing hooks to DHCP, TFTP, and HTTP with the goal of making end-to-end traces work.

## Why is this needed

Brings all the components together to carry a trace through an OS install.

Should be a no-op for anyone not enabling OpenTelemetry. For those who do use otel, traces will be emitted to the OTLP endpoint they choose.

## Checklist:

I have:

- [ ] added unit or e2e tests


tobert:
Whoops I branched off #183 and it took everything. Will wait to PR until that's merged and I can rebase.

PR #206, opened 2021-09-17
installers: remove f3 plan specific tty

dustinmiller:
Found a method to ensure those plans were on ttyS1 and so we won't need 
to adjust to ttyS0 anymore.

Signed-off-by: Dustin Miller <dustin@packet.com>

## Description

Remove plan specific actions for console tty settings

## Why is this needed

More defaults and less Metal specific things is good

## How Has This Been Tested?

Just in theory


mmlb (approved):

jacobweinstock (approved):
Thanks, @dustinmiller !

PR #207, opened 2021-09-24
kickstart esxi: Remove the -s option from wget

dlaube:
## Description

The VMware installer was previously relying on wget's "-s" or spider option to determine whether or not a particular URL was accessible while handling the customdata portion of the kickstart. This option is not valid on the version of wget which ships with current closed-source ESXi installers.

## Why is this needed

This change fixes a bug related to the customdata logic responsible for executing the kickstart postinstall scripts and kickstart firstboot install scripts provided via customdata.

Fixes: #

## How Has This Been Tested?
CI testing

Fixes customdata bug and unblocks further VMware VCF configuration.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/installers/vmware/kickstart-esxi.go
+++ b/installers/vmware/kickstart-esxi.go
@@ -231,17 +231,10 @@ fi
 # Kickstart firstboot supplemental config URL
 if [ "$kickstartfburl" != "null" ]; then
 	echo "Using supplemental kickstart firstboot URL: $kickstartfburl"
-	if wget -q $kickstartfburl; then
-		wget -q $kickstartfburl -O /tmp/ks-firstboot-sup.sh
-		echo "========Begin execution of supplemental firstboot kickstart"
-		chmod +x /tmp/ks-firstboot-sup.sh && /tmp/ks-firstboot-sup.sh
-		echo "========End execution of supplemental firstboot kickstart"
-	else
-		echo "ERROR: Custom kickstart firstboot URL is NOT accessible!"
-		exit 1
-	fi
-else
-	echo "Skipping supplemental kickstart firstboot URL"
+	wget -q $kickstartfburl -O /tmp/ks-firstboot-sup.sh
+	echo "========Begin execution of supplemental firstboot kickstart"
+	chmod +x /tmp/ks-firstboot-sup.sh && /tmp/ks-firstboot-sup.sh
+	echo "========End execution of supplemental firstboot kickstart"
 fi
 
 # Kickstart firstboot supplemental shell commands
@@ -308,18 +301,13 @@ kickstartpishellcmd=$(custom_data "['kickstart']['postinstall_shell_cmd']")
 
 # Kickstart postinstall supplemental config URL
 if [ "$kickstartpiurl" != "null" ]; then
-        echo "Using supplemental kickstart postinstall URL: $kickstartpiurl"
-        if wget -q $kickstartpiurl; then
-                wget -q $kickstartpiurl -O /tmp/ks-postinstall-sup.sh
-                echo "========Begin execution of supplemental postinstall kickstart"
-                chmod +x /tmp/ks-postinstall-sup.sh && /tmp/ks-postinstall-sup.sh
-                echo "========End execution of supplemental postinstall kickstart"
-        else
-                echo "ERROR: Custom kickstart postinstall URL is NOT accessible!"
-                exit 1
-        fi
+	echo "Using supplemental kickstart postinstall URL: $kickstartpiurl"
+	wget -q $kickstartpiurl -O /tmp/ks-postinstall-sup.sh
+	echo "========Begin execution of supplemental postinstall kickstart"
+	chmod +x /tmp/ks-postinstall-sup.sh && /tmp/ks-postinstall-sup.sh
+	echo "========End execution of supplemental postinstall kickstart"
 else
-        echo "Skipping supplemental kickstart postinstall URL"
+	echo "Skipping supplemental kickstart postinstall URL"
 fi
 
 # Kickstart postinstall supplemental shell commands
--- a/installers/vmware/kickstart-esxi.go
+++ b/installers/vmware/kickstart-esxi.go
@@ -231,10 +231,16 @@ fi
 # Kickstart firstboot supplemental config URL
 if [ "$kickstartfburl" != "null" ]; then
 	echo "Using supplemental kickstart firstboot URL: $kickstartfburl"
-	wget -q $kickstartfburl -O /tmp/ks-firstboot-sup.sh
-	echo "========Begin execution of supplemental firstboot kickstart"
-	chmod +x /tmp/ks-firstboot-sup.sh && /tmp/ks-firstboot-sup.sh
-	echo "========End execution of supplemental firstboot kickstart"
+	if wget -q "$kickstartfburl" -O /tmp/ks-firstboot-sup.sh
+		echo "========Begin execution of supplemental firstboot kickstart"
+		chmod +x /tmp/ks-firstboot-sup.sh && /tmp/ks-firstboot-sup.sh
+		echo "========End execution of supplemental firstboot kickstart"
+	else
+		echo "ERROR: Custom kickstart firstboot URL '$kickstartfburl' is NOT accessible!"
+		exit 1
+	fi
+else
+	echo "Skipping supplemental kickstart firstboot URL"
 fi
 
 # Kickstart firstboot supplemental shell commands
--- a/installers/vmware/kickstart-esxi.go
+++ b/installers/vmware/kickstart-esxi.go
@@ -231,7 +231,7 @@ fi
 # Kickstart firstboot supplemental config URL
 if [ "$kickstartfburl" != "null" ]; then
 	echo "Using supplemental kickstart firstboot URL: $kickstartfburl"
-	if wget -q -s $kickstartfburl; then
+	if wget -q $kickstartfburl; then
 		wget -q $kickstartfburl -O /tmp/ks-firstboot-sup.sh
 		echo "========Begin execution of supplemental firstboot kickstart"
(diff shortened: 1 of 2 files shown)


tstromberg (changes requested):
Can you change the PR title to something meaningful to display in our release notes? A 1-line change of what change is introduced would be hugely helpful here. Something like `kickstart esxi: Remove the -s option from wget`

The PR description needs some sort of "Why" this change was necessary, and what was the best way to go about fixing it. Trying to discover what this `-s` option did is surprisingly difficult without `wget` being installed on my machine.

Also, seeing an SE-356 moniker is not helpful to the majority of the folks who do not have access to your issue tracking system.

dlaube:
@tstromberg Updated per your recommendations.

installers/vmware/kickstart-esxi.go
```diff
@@ -231,7 +231,7 @@ fi
 # Kickstart firstboot supplemental config URL
 if [ "$kickstartfburl" != "null" ]; then
 	echo "Using supplemental kickstart firstboot URL: $kickstartfburl"
-	if wget -q -s $kickstartfburl; then
+	if wget -q $kickstartfburl; then
 		wget -q $kickstartfburl -O /tmp/ks-firstboot-sup.sh
```
mmlb, line 235: Why download twice? This could just be
```suggestion
	if wget -q $kickstartfburl -O /tmp/ks-firstboot-sup.sh; then
```

installers/vmware/kickstart-esxi.go
```diff
@@ -309,7 +309,7 @@ kickstartpishellcmd=$(custom_data "['kickstart']['postinstall_shell_cmd']")
 # Kickstart postinstall supplemental config URL
 if [ "$kickstartpiurl" != "null" ]; then
         echo "Using supplemental kickstart postinstall URL: $kickstartpiurl"
-        if wget -q -s $kickstartpiurl; then
+        if wget -q $kickstartpiurl; then
                 wget -q $kickstartpiurl -O /tmp/ks-postinstall-sup.sh
```
mmlb, line 313:
```suggestion
        if wget -q $kickstartpiurl -O /tmp/ks-postinstall-sup.sh; then
```

mmlb (changes requested):

installers/vmware/kickstart-esxi.go
dlaube: SGTM

installers/vmware/kickstart-esxi.go
dlaube: Done :)

installers/vmware/kickstart-esxi.go
```diff
@@ -231,17 +231,10 @@ fi
...
-		echo "========Begin execution of supplemental firstboot kickstart"
-		chmod +x /tmp/ks-firstboot-sup.sh && /tmp/ks-firstboot-sup.sh
-		echo "========End execution of supplemental firstboot kickstart"
-	else
-		echo "ERROR: Custom kickstart firstboot URL is NOT accessible!"
-		exit 1
-	fi
-else
-	echo "Skipping supplemental kickstart firstboot URL"
+	wget -q $kickstartfburl -O /tmp/ks-firstboot-sup.sh
```
mmlb, line 234: hmm but now you got rid of the handling of `wget` return code, that seems worse right?

installers/vmware/kickstart-esxi.go
```diff
@@ -308,18 +301,13 @@ kickstartpishellcmd=$(custom_data "['kickstart']['postinstall_shell_cmd']")
 
 # Kickstart postinstall supplemental config URL
 if [ "$kickstartpiurl" != "null" ]; then
```
displague, line 303: Should we also ignore empty urls?

dlaube: We do want to check to make sure we dont try to wget when there isn't a URL provided.

installers/vmware/kickstart-esxi.go
dlaube: Possibly :) Should we throw that if check back in then?

mmlb: yeah the if was good, but we didn't need to do the download twice. Could just do the download and branch on success/failure like:
```if ! wget --fail -o ...; then
# handle wget failed.
...
```

dlaube: Unfortunately option "--fail" is either too new or is not supported in the vmware installer. Looking at this further, even though the "-s" option was dropped its still supposed to support "--spider"

installers/vmware/kickstart-esxi.go
gemagomez: We shouldn't be ignoring empty URLs no.

installers/vmware/kickstart-esxi.go
tstromberg: Unless I'm missing something, `--fail` isn't necessary to do the logic check, wget emits an exit code either way. Here is my personal recommendation:

```shell
wget -q "${kickstartfburl}" -O /tmp/ks-firstboot-sup.sh \
  || { echo  "ERROR: Custom kickstart firstboot URL '${kickstartfburl}' is NOT accessible"; exit 1;}
```

(do note the trailing semi-colon)

alternatively:

```shell
if ! wget -q "${kickstartfburl}" -O /tmp/ks-firstboot-sup.sh; then
  echo  "ERROR: Custom kickstart firstboot URL '${kickstartfburl}' is NOT accessible"
  exit 1
fi
```

I've locally tested both of these options and they appear to function properly. Please let me know if you need any help getting this sorted out.

mmlb: Ah sorry, my familiarity with wget is less than with curl. Does wget return an error if we get a 404 or even 5xx error?

tstromberg: Yes, it does in my testing, generally 8:

```
EXIT STATUS
       Wget may return one of several error codes if it encounters problems.

       0   No problems occurred.
       1   Generic error code.
       2   Parse error---for instance, when parsing command-line options, the .wgetrc or .netrc...
       3   File I/O error.
       4   Network failure.
       5   SSL verification failure.
       6   Username/password authentication failure.
       7   Protocol errors.
       8   Server issued an error response.
```

mmlb: Ah nice thanks for looking that up for me. I should have just RTFM :(.

installers/vmware/kickstart-esxi.go
```diff
@@ -308,18 +301,13 @@ kickstartpishellcmd=$(custom_data "['kickstart']['postinstall_shell_cmd']")
...
-                echo "========End execution of supplemental postinstall kickstart"
-        else
-                echo "ERROR: Custom kickstart postinstall URL is NOT accessible!"
-                exit 1
-        fi
+	echo "Using supplemental kickstart postinstall URL: $kickstartpiurl"
+	wget -q $kickstartpiurl -O /tmp/ks-postinstall-sup.sh
+	echo "========Begin execution of supplemental postinstall kickstart"
+	chmod +x /tmp/ks-postinstall-sup.sh && /tmp/ks-postinstall-sup.sh
+	echo "========End execution of supplemental postinstall kickstart"
```
mmlb, line 308: Ditto. Seems like we went off the track a little here, can you apply this suggestion instead?

```suggestion
	echo "Using supplemental kickstart postinstall URL: $kickstartpiurl"
	if wget -q "$kickstartpiurl" -O /tmp/ks-postinstall-sup.sh
		echo "========Begin execution of supplemental postinstall kickstart"
		chmod +x /tmp/ks-postinstall-sup.sh && /tmp/ks-postinstall-sup.sh
		echo "========End execution of supplemental postinstall kickstart"
	else
		echo "ERROR: Custom kickstart postinstall URL '$kickstartpiurl' is NOT accessible!"
		exit 1
	fi
```

installers/vmware/kickstart-esxi.go
```diff
@@ -231,17 +231,10 @@ fi
...
-	else
-		echo "ERROR: Custom kickstart firstboot URL is NOT accessible!"
-		exit 1
-	fi
-else
-	echo "Skipping supplemental kickstart firstboot URL"
+	wget -q $kickstartfburl -O /tmp/ks-firstboot-sup.sh
+	echo "========Begin execution of supplemental firstboot kickstart"
+	chmod +x /tmp/ks-firstboot-sup.sh && /tmp/ks-firstboot-sup.sh
+	echo "========End execution of supplemental firstboot kickstart"
```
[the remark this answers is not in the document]
mmlb: This gets us back to what the original code looked like with error handling but avoids downloading the files twice.

installers/vmware/kickstart-esxi.go
```diff
@@ -308,18 +307,17 @@ kickstartpishellcmd=$(custom_data "['kickstart']['postinstall_shell_cmd']")
...
-                wget -q $kickstartpiurl -O /tmp/ks-postinstall-sup.sh
-                echo "========Begin execution of supplemental postinstall kickstart"
-                chmod +x /tmp/ks-postinstall-sup.sh && /tmp/ks-postinstall-sup.sh
-                echo "========End execution of supplemental postinstall kickstart"
-        else
-                echo "ERROR: Custom kickstart postinstall URL is NOT accessible!"
-                exit 1
-        fi
+	echo "Using supplemental kickstart postinstall URL: $kickstartpiurl"
+	if wget -q "$kickstartpiurl" -O /tmp/ks-postinstall-sup.sh
```
tstromberg, line 311: this line is going to need a "; then" suffix.

mmlb: gah! copy/paste fail in github editor :(

dlaube: good catch! updated

mmlb (approved):
lgtm! sorry for the confusion earlier.

tstromberg (approved):

PR #208, opened 2021-09-28
installers: apply bootdrive hint for w plans

dustinmiller:
Signed-off-by: Dustin Miller <dustin@packet.com>

## Description

Lets use the the BootDriveHint method for plans which start with `w` as well

## Why is this needed

Picking the wrong disk has negative impacts on provisioning

## How Has This Been Tested?

Just theory

### DIFF
--- a/installers/vmware/kickstart-esxi.go
+++ b/installers/vmware/kickstart-esxi.go
@@ -388,6 +388,9 @@ func determineDisk(j job.Job) string {
 	if j.BootDriveHint() != "" && strings.HasPrefix(j.PlanSlug(), "s") {
 		return "--firstdisk=" + j.BootDriveHint()
 	}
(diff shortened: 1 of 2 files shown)


tstromberg (approved):
Is there any indication that we are unable to use the boot drive hint for all plans at this point?

As we're trying to eliminate all of the plan slugs included in the code base, it would be nice if we could just remove the case statement below here (in this or a future PR).

dustinmiller:
> Is there any indication that we are unable to use the boot drive hint for all plans at this point?
> 
> As we're trying to eliminate all of the plan slugs included in the code base, it would be nice if we could just remove the case statement below here (in this or a future PR).

This will expand the scope of the boot drive hinting greatly and should facilitate getting the remainder of the plan based code for firstdisk removed, assuming it's working as intended

PR #217, opened 2021-10-12
Revert "Refactor setPXEFilename to be less complicated"

jacobweinstock:
Reverts tinkerbell/boots#149

Testing this at Equinix Metal appears to not be compatible with some ARM servers. My desire is to revert this and then do some more testings to figure out why.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,2 +1,2 @@
-has nix && use nix
-PATH_add bin
+which nix &>/dev/null && use nix
+unset GOPATH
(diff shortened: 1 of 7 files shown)


mmlb:
This is the same thing as what is causing #225 . The fw/nics are ignoring boot file being offered if PXEClient is set. I was able to verify this while debugging #225. Both that PR and this have similar code in common https://github.com/tinkerbell/boots/pull/149/files#diff-757d4edc379c9bd53273363379744ab01fd32b8d4977cb13929c42b76f526a93R97-R103 and https://github.com/tinkerbell/boots/pull/216/files#diff-757d4edc379c9bd53273363379744ab01fd32b8d4977cb13929c42b76f526a93R140-R145

PR #221, opened 2021-11-16
Enable PCI Pass-through and IOMMU as a default kernel parameter

truongmd:
## Description

Enabled PCI Pass-through mode and enables the IOMMU by default

## Why is this needed

Needed to enable ISO support

No expected user impact

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/installers/osie/main.go
+++ b/installers/osie/main.go
@@ -119,6 +119,8 @@ func kernelParams(ctx context.Context, action, state string, j job.Job, s ipxe.S
 	switch j.PlanSlug() {
 	case "c2.large.arm", "c2.large.anbox":
 		s.Args("iommu.passthrough=1")
+        default:
+                s.Args("intel_iommu=on amd_iommu=on iommu=pt")
 	}
 
 	if action == "install" {
--- a/installers/osie/main.go
+++ b/installers/osie/main.go
@@ -117,10 +117,10 @@ func kernelParams(ctx context.Context, action, state string, j job.Job, s ipxe.S
 	s.Args("facility=" + j.FacilityCode())
 
 	switch j.PlanSlug() {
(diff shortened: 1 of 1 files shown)


installers/osie/main.go
```diff
@@ -119,6 +119,8 @@ func kernelParams(ctx context.Context, action, state string, j job.Job, s ipxe.S
 	switch j.PlanSlug() {
 	case "c2.large.arm", "c2.large.anbox":
 		s.Args("iommu.passthrough=1")
+        default:
```
zevweiss, line 122: Indentation looks mismatched (spaces vs. tabs); run `gofmt` to fix I'd expect?

installers/osie/main.go
```diff
@@ -119,6 +119,8 @@ func kernelParams(ctx context.Context, action, state string, j job.Job, s ipxe.S
...
+                s.Args("intel_iommu=on amd_iommu=on iommu=pt")
```
zevweiss, line 123: I don't think `on` is a recognized value for `amd_iommu` (see https://www.kernel.org/doc/Documentation/admin-guide/kernel-parameters.txt), so we should probably take that one out.

zevweiss (changes requested):

installers/osie/main.go
```diff
@@ -119,6 +119,8 @@ func kernelParams(ctx context.Context, action, state string, j job.Job, s ipxe.S
 	switch j.PlanSlug() {
 	case "c2.large.arm", "c2.large.anbox":
```
enkelprifti98, line 120: Could we add the recently released "c3.large.arm" here?

truongmd: Added in a following commit

PR #223, opened 2021-11-18
Get rid of ipxe/ names in the binaries offered up for tftp and http boot

mmlb:
## Description

Changes the binary names from ipxe/$name to just $name. Fixes #223

## Why is this needed

The filenames recently changed as part of #216 which was a great feature. Unfortunately old UEFI nic fw does not agree. When tested against some of the EM uefi machines (c2.large.arm) they failed to boot. The machines did not even attempt to tftp the file.

## How Has This Been Tested?

CI only so far, will be tested against the previously failing machines tomorrow.

Not a real big impact, just that the file names won't be so easy to find when looking at tftp or http boot logs.

## Checklist:

I have:

- [x] added unit or e2e tests


tstromberg:
What work is required to push this PR past the draft state?

nshalman:
>Not a real big impact, just that the file names won't be so easy to find when looking at tftp or http boot logs.

Could we enrich the log messages to make them easier to search?

mmlb:
> What work is required to push this PR past the draft state?

I'm testing this using the sandbox to spin up arm machines (based off of https://github.com/mmlb/sandbox/tree/fix-terraform-and-add-arm64-support) I only just got it to work around EoD last friday. I was able to reproduce the issue in my sandbox setup but this branch did not actually work either. I've yet to examine a tcpdump of good vs bad to ensure things are as expected. So still need some more testing/debugging of this branch.

mmlb:
This is not the issue with #225 and is unnecessary

PR #224, opened 2021-11-23
update ipxe commit to include URL error fix

deitch:
## Description

There is a bug in ipxe wherein it chokes on certain URL formats. This was particularly noted in some of how GitHub distributes assets in changes they made over the last month, which meant that iPXE scripts that referenced assets on GitHub would fail.

[This patch](https://github.com/ipxe/ipxe/pull/516) fixes it in iPXE; this PR bumps boots's version of iPXE to the commit that includes the patch.

There should be none


nshalman (approved):
LGTM

PR #226, opened 2021-12-01
Inband traceparent propagation for cacher

tobert:
## Description

When a system provision is kicked off, a hardware record is written to Cacher and that's where the synchronous chain ends usually, because Boots is usually triggered by a DHCP or IP request from a system under its management. This allows for in-band transmission of traceparent strings so the upstream of Cacher can inject it in the record and then boots can create the span links so it can all be tied together in observability tools.

## Why is this needed

Accepts OpenTelemetry / w3c traceparent in-band in hardware records and sets an otel span link on Boots spans so they can be tracked in observability tools.

## How Has This Been Tested?

Tests pass, CI of course. I have run tests locally using standalone mode with and without a traceparent and it seems to work fine and the traces look good.

The trace links might not work because I couldn't figure out how to validate them, but even if they don't work they'll be harmless and we can remove them later.

No migration necessary. Should not impact anything but OpenTelemetry.


tobert:
Gonna think on this some more... I might be able to do better than the link but it'll mean more code churn.

mmlb:
lgtm so far, curious to see the alternative you're thinking about. Could you throw in some simple unit tests too please?

tobert:
> lgtm so far, curious to see the alternative you're thinking about.

As we discussed elsewhere, the alternative is only link the initial DHCP span, and then switch to the traceparent from API from there on.

> Could you throw in some simple unit tests too please?

I get that codecov is complaining but the code I added doesn't do anything unit tests can address and I think its calculations are bogus here. Or, do you unit test your metrics and logs already? This is in the same class.

job/fetch.go
```diff
@@ -12,16 +12,16 @@ var (
...
 
-func discoverHardwareFromDHCP(ctx context.Context, mac net.HardwareAddr, giaddr net.IP, circuitID string) (packet.Discovery, error) {
+func discoverHardwareFromDHCP(ctx context.Context, mac net.HardwareAddr, giaddr net.IP, circuitID string) (context.Context, packet.Discovery, error) {
 	fetch := func() (interface{}, error) {
 		return client.DiscoverHardwareFromDHCP(ctx, mac, giaddr, circuitID)
 	}
 	v, err := servers.Do(mac.String(), fetch)
 	if err != nil {
-		return nil, err
+		return ctx, nil, err
```
mmlb, line 21: can we return nil for the context instead, it just feels simpler. The caller won't have to worry about doing anything with the context being returned and can just keep it out of mind.

job/job.go
```diff
@@ -90,47 +89,37 @@ func HasActiveWorkflow(ctx context.Context, hwID packet.HardwareID) (bool, error
...
-		trace.WithAttributes(attribute.String("CircuitID", circuitID)),
-	)
-
-	d, err := discoverHardwareFromDHCP(ctx, mac, giaddr, circuitID)
+	ctx, d, err := discoverHardwareFromDHCP(ctx, mac, giaddr, circuitID)
 	if err != nil {
-		span.SetStatus(codes.Error, err.Error())
-
-		return Job{}, errors.WithMessage(err, "discover from dhcp message")
+		return ctx, Job{}, errors.WithMessage(err, "discover from dhcp message")
```
mmlb, line 100: ditto, lets return nil here for similar reasons as in fetch.go. Here and in the rest of the function's error branches.

packet/models_test.go
```diff
@@ -434,7 +440,8 @@ var cacherTests = map[string]struct {
 			Gateway: net.ParseIP("10.250.142.1"),
 			Netmask: net.ParseIP("255.255.255.0"),
 		},
-		json: discovered,
+		traceparent: "",
```
mmlb, line 443: no need to set traceparent to `""` thats what it'll get by default anyway. Setting to default just adds visual noise, I think its best to just have the "set bits" or "things being tested" specified. Can you please remove.

cmd/boots/tftp.go
```diff
@@ -68,21 +68,21 @@ func (t tftpHandler) ReadFile(c tftp.Conn, filename string) (tftp.ReadCloser, er
 		l.Info("client requested filename '", filename, "' with a traceparent attached and has been shortened to '", shortfile, "'")
 		filename = shortfile
 	}
+
 	tracer := otel.Tracer("TFTP")
-	ctx, span := tracer.Start(ctx, "TFTP get",
+	// save the context in a different var so it can be used to create links later
+	rctx, span := tracer.Start(ctx, "TFTP get",
```
mmlb, line 74: I didn't notice this last time, but it would be nice to make this string match the tftp protocol's wording I think. Could you change `TFTP get` to `TFTP Read` instead? Not a big deal if you'd rather not we can address separately though.

cmd/boots/tftp.go
```diff
@@ -96,15 +96,17 @@ func (t tftpHandler) ReadFile(c tftp.Conn, filename string) (tftp.ReadCloser, er
...
-		span.End()
 
 		return serveFakeReader(l, filename)
 	}
 
-	span.SetStatus(codes.Ok, filename)
-	span.End()
+	// this span is after fetching the hw data so it will be attached to the upstream trace id
+	// we also link the spans here so they can be tied together in the tracing UI
+	ctx, uSpan := tracer.Start(ctx, "TFTP open", trace.WithAttributes(attribute.String("filename", filename)))
```
mmlb, line 105: nit: Should be `TFTP Open` to match the case used for DHCP

packet/models_cacher.go
```diff
@@ -41,6 +41,7 @@ type HardwareCacher struct {
 	ServicesVersion   ServicesVersion `json:"services"`
 	Instance          *Instance       `json:"instance"`
 	ProvisionerEngine string          `json:"provisioner_engine"`
+	Traceparent       string          `json:"traceparent"`
```
mmlb, line 44: is `traceparent` set it stone? It kind of bugs me that its not `trace_parent` in the json and `TraceParent` in the struct. Its not a big deal... just a minor annoyance really but if we can change it that would be great.

mmlb:
> > lgtm so far, curious to see the alternative you're thinking about.
> 
> As we discussed elsewhere, the alternative is only link the initial DHCP span, and then switch to the traceparent from cacher/upstream from there on.
> 
> > Could you throw in some simple unit tests too please?
> 
> I get that codecov is complaining but the code I added doesn't do anything unit tests can address and I think its calculations are bogus here. Or, do you unit test your metrics and logs already? This is in the same class.

I don't recall rn but I think when this was just mostly the GetTracparent code having something checking that it works as expected in cacher and tink modes would have quieted it down. Looks like you did that already and codecov is happy, so I'm happy too :D

cmd/boots/tftp.go
tobert: > 

will do, I was thinking about refactoring all these names now that we know more about how otel fits into the code

for now I'll update the bits in this PR and follow up later with another that refactors the rest into some normalcy, probably along the lines of the otel semantic conventions https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md

cmd/boots/tftp.go
tobert: gonna switch these over to the otel semantic recommendations. for this one, "tftp/ServeTFTP" since that's the method the span is wrapping

job/fetch.go
tobert: > can we return nil for the context instead, it just feels simpler. The caller won't have to worry about doing anything with the context being returned and can just keep it out of mind.

I do not think we should do that. While boots might usually be done with a context when there is an error, that's not a reason to break the context chain. Especially so when we're doing otel things through context. Nobody ever checks ctx for nil, and they shouldn't have to.

packet/models_cacher.go
tobert: Unfortunately, yes it is set in stone. It's hard-coded in the otel spec and libraries, and they make it _really_ hard to mess with, without adding a lot of ugly boilerplate. We're using `traceparent` and `Traceparent` all over the place and I don't think it should be camelcased.

packet/models_test.go
tobert: I was just trying to be clear but okay.

job/job.go
tobert: > ditto, lets return nil here for similar reasons as in fetch.go. Here and in the rest of the function's error branches.

As I mentioned, I believe this is the wrong move. It's your code so if you really want this change lmk and I'll do it anyways.

tobert:
@mmlb Could I get you to take another look please?

job/fetch.go
mmlb: Actually I don't see why we need to return the context here at all. Its never being mutated and thus the return is only confusing right?

job/job.go
mmlb: Yep I see it that in this function the `ctx` is likely different even in the error case (if `j.setup()` is what returns an error) which I hadn't noticed before so sounds fine to not return `nil` as that will be confusing to callers.

cmd/boots/dhcp.go
```diff
@@ -86,7 +86,9 @@ func (d dhcpHandler) serveDHCP(w dhcp4.ReplyWriter, req *dhcp4.Packet) {
 		trace.WithAttributes(attribute.String("CircuitID", circuitID)),
 	)
 
-	j, err := job.CreateFromDHCP(ctx, mac, gi, circuitID)
+	// otel: if the hw record this retrieves has a traceparent, the job code will make a span
+	// link and swap out the traceparent in the context from here on
```
mmlb, line 90: This comment actually belongs in `job.CreateFrom*`'s doc string instead of (or in addition to) here.

job/job.go
```diff
@@ -90,47 +89,37 @@ func HasActiveWorkflow(ctx context.Context, hwID packet.HardwareID) (bool, error
 }
 
 // CreateFromDHCP looks up hardware using the MAC from cacher to create a job
```
mmlb, line 91: should document that the returned `ctx` is likely to be modified (even in error case) and should thus overtake the one used as the arg.

job/job.go
```diff
@@ -90,47 +89,37 @@ func HasActiveWorkflow(ctx context.Context, hwID packet.HardwareID) (bool, error
...
 	if err != nil {
-		span.SetStatus(codes.Error, err.Error())
 		j = Job{} // return an empty job on error
 	}
 
-	return j, err
+	return ctx, j, err
 }
 
 // CreateFromRemoteAddr looks up hardware using the IP from cacher to create a job
```
mmlb, line 111: ditto re returned ctx

job/job.go
```diff
@@ -90,47 +89,37 @@ func HasActiveWorkflow(ctx context.Context, hwID packet.HardwareID) (bool, error
...
 	host, _, err := net.SplitHostPort(ip)
 	if err != nil {
-		return Job{}, errors.Wrap(err, "splitting host:ip")
+		return ctx, Job{}, errors.Wrap(err, "splitting host:ip")
 	}
 
 	return CreateFromIP(ctx, net.ParseIP(host))
 }
 
 // CreateFromIP looksup hardware using the IP from cacher to create a job
```
mmlb, line 121: and here

job/job.go
```diff
@@ -176,11 +165,19 @@ func (j Job) MarkDeviceActive(ctx context.Context) {
 	}
 }
 
-func (j *Job) setup(d packet.Discovery) error {
+func (j *Job) setup(ctx context.Context, d packet.Discovery) (context.Context, error) {
```
mmlb, line 168: definitely need to document the returned ctx here.

packet/models_test.go
mmlb: These still need to be dropped

mmlb (changes requested):

mmlb:
> @mmlb Could I get you to take another look please?

sorry for going awol for a bit here :|

job/fetch.go
tobert: Yep, I traced back through the code paths and don't remember why I even added it. Probably started thinking I'd need it then found the other path in job.setup. Commit coming soon to remove.

job/job.go
tobert: added

tobert:
@mmlb no worries about the time, I still need to circle back to the API side of things.

I believe I have addressed all of your comments.

packet/models_test.go
tobert: Gone in the diff now :)

job/job.go
mmlb: thanks

tftp/tftp.go
```diff
@@ -1,12 +1,17 @@
 package tftp
 
 import (
+	"context"
+	_ "embed"
```
mmlb: hmm this is surprising, this shouldn't be necessary here.

mmlb: Yep not necessary but also not cleaned up automagically so I did so.

mmlb (approved):

mmlb:
Fixed the unnecessary `_ "embed"` import and rebased onto main + squashed in some commits to avoid history churn. Should be merged once tests pass.

mmlb:
> @mmlb no worries about the time, I still need to circle back to the API side of things.

Is this good to merge or do we need to wait still?
 
> I believe I have addressed all of your comments.

Yep!

tobert:
> > @mmlb no worries about the time, I still need to circle back to the API side of things.
> 
> Is this good to merge or do we need to wait still?

Good to merge. Thanks!

tobert:
> 



> Fixed the unnecessary `_ "embed"` import and rebased onto main + squashed in some commits to avoid history churn. Should be merged once tests pass.

That came in when I merged master and go mod tidy didn't complain and it looked harmless so I let it through. Dunno how it showed up 🤷🏻‍♀️

mmlb:
> > 
> 
> > Fixed the unnecessary `_ "embed"` import and rebased onto main + squashed in some commits to avoid history churn. Should be merged once tests pass.
> 
> That came in when I merged master and go mod tidy didn't complain and it looked harmless so I let it through. Dunno how it showed up 🤷🏻‍♀️

It was around from when tftp/ package had the ipxe binaries within it but latest main moved things around out of there I guess it stuck around any you pulled it in when merging main in. :shrug:

PR #228, opened 2021-12-08
Split and fix http boot for c2.large.arm

### LINKED ISSUE #210: Add support for UEFI HTTP Boot
## Expected Behaviour

I expected to be able to use [UEFI HTTP Boot](https://github.com/tianocore/tianocore.github.io/wiki/HTTP-Boot) to boot a machine.

## Current Behaviour

Only PXE Boot is supported.

## Possible Solution

I've got dnsmasq serving HTTP Boot at:

https://github.com/rgl/talos-vagrant/blob/6ce54f83e4066bc60415d697f5c4a7a6cd5ac561/provision-dnsmasq.sh#L53-L59

### LINKED ISSUE #225: Investigate PXE issue with Lenovo ThinkSystem HR330A (Ampere eMAG)
Apparently, https://github.com/tinkerbell/boots/pull/216 may have broken PXE boot on this hardware. I'm not sure what BIOS version, however.

@mmlb mentioned this on chat: 

> ... the last bunch of dhcps ... where the filename is under ipxe/ and nothing back from the nic/uefi. I was on console and the machine went almost straight to uefi shell.

#223 may fix this, but if we can confirm the problem we should probably just roll-back #216 and bring it back without the path changes.

mmlb:
## Description

This is a revert of the revert that occurred in #217 + fix that caused the original revert.
There's also removal of what is effectively useless/dead code (isPXEClient/pxeClient var/handling).
I also removed the leading `/` from `/nonexistent` because a `/` is already prefixed and `//nonexistent` has bothered me for years.

## Why is this needed

#216 was a great contribution from @rgl, unfortunately we ran into issues on some of our Ampere eMAG systems (#225).
Turns out that the eMAG boot firmware just straight up ignores filename values if the response packet is marked as `PXEClient`.
#216 changed the logic subtly by setting the response packet's class as `HTTPClient` if http was supported by the booting machine, and otherwise setting `PXEClient`. Before #216 we would set `PXEClient` only for the dhcp responses that were going out to our tftp'd build of iPXE, the initial filename: response was `PXEClient`-less.

Once I put back the only `PXEClient` for our iPXE everything worked out fine again (tested using sandbox), but then I realized that now pxeClient was likely useless because I suspected iPXE did not care/require it. I tested this too and found that iPXE is happy to tftp a file provided by a `PXEClient`-less DHCP response. So why keep code around that is effectively useless? Less code is better than any code so that caused me to write up d1878bddb9dcbdde848cbbd485e5cb30622105a5.

Fixes #210
Fixes #225

This debugging also opens up the way to bring back #149 as that had to be reverted due to the same "always set `PXEClient`" issue that I was not equipped to debug easily/quickly then.

## How Has This Been Tested?

Lots of testing on EM machines using github.com/tinkerbell/sandbox's terraform support (I plan to clean up and PR https://github.com/mmlb/sandbox/tree/fix-terraform-and-add-arm64-support at some point).

Back to faster/better iPXE binary downloads by utilizing TCP/HTTP instead of UDP/TFTP.


mmlb:
@rgl I'd love you to give this a go on your machines too please.

rgl:
@mmlb it works here too! thanks for looking into this :-)

nshalman (approved):

Issue #213, opened 2021-10-07
Move ipxe to a dedicated repository?


rgl:
Currently ipxe is built in this repository, that makes the build of boots take longer and is more complex.

What do you think about:

1. moving the ipxe build to the boots-ipxe repository.
2. build ipxe in a GitHub Actions inside a docker container using buildx from a ubuntu 20.04 image.
3. publish an artifact for each arch as a GitHub Release.
4. modify the boots build to use the GitHub Release.

This would make things simpler from the boots build perspective and for someone that is trying to understand and build boots.

mmlb:
Hey @rgl this sounds like a great suggestion, I like it. There really is no need to be wasting electricity building ipxe over and over again.

I have just 2 push backs.
1. I don't see the benefit of building using buildx over make
2. I'd rather stick with nix's pinned inputs over ubuntu.

Issue #130, opened 2021-02-17
boots doesn't compile using `go build`: undefined: ipxe.MustAsset


Smithx10:
[root@73a5a635-85b2-4a26-cd19-bd3e89bc9c36 boots]$ go build -v
github.com/tinkerbell/boots/tftp
# github.com/tinkerbell/boots/tftp
tftp/tftp.go:15:20: undefined: ipxe.MustAsset
tftp/tftp.go:16:20: undefined: ipxe.MustAsset
tftp/tftp.go:17:20: undefined: ipxe.MustAsset
tftp/tftp.go:18:20: undefined: ipxe.MustAsset

mmlb:
Hey @Smithx10 can you try with `make`, that should work. I've been noodling a way to still be `go build`able with just degraded functionality.

Smithx10:
@mmlb  Thanks for the quick response,

I'm attempting to compile for amd64.

```
arch@9d46dd2d-1178-ce83-cbc4-d396e4a24060 ~/g/boots ❯❯❯ make boots-linux-amd64                                                                                                                 ✘ 130 master
t=bin-arm64-efi/snp.efi
rm -rf ipxe/ipxe/build/bin-arm64-efi
mkdir -p ipxe/ipxe/build/bin-arm64-efi
tar -xzf ipxe/ipxe/build/a5fb41873dc1dcce5dcc817bf53a0649c01f8cac.tar.gz -C ipxe/ipxe/build/bin-arm64-efi
cp ipxe/ipxe/colour.h ipxe/ipxe/common.h ipxe/ipxe/console.h ipxe/ipxe/general.efi.h ipxe/ipxe/general.undionly.h ipxe/ipxe/isa.h ipxe/ipxe/build/bin-arm64-efi
cd ipxe/ipxe/build/bin-arm64-efi && ../../build.sh $t a5fb41873dc1dcce5dcc817bf53a0649c01f8cac
+ case $build in
+ rm ipxe-a5fb41873dc1dcce5dcc817bf53a0649c01f8cac/src/config/local/isa.h
+ cp ipxe-a5fb41873dc1dcce5dcc817bf53a0649c01f8cac/src/config/local/general.efi.h ipxe-a5fb41873dc1dcce5dcc817bf53a0649c01f8cac/src/config/local/general.h
[...]

Issue #66, opened 2020-08-18
Make TFTP session timeout tunable (defaults to 3 seconds)


lukaskalvenas:
It would be nice to have the possibility to edit the TFTP session timeout for each worker server. Currently, if the worker server does not respond within 3-6 seconds, TFTP times out and the operating system installation ultimately fails. 

## Current Behaviour
CLIENT MAC ADDR: $MAC  GUID: 00000000 0000 0000 0000 AC1F6BA50120
CLIENT IP: xxx.xxx.xxx.xxx  MASK: yyy.yyy.yyy.yyy  DHCP IP: zzz.zzz.zzz.zzz
GATEWAY IP: aaa.aaa.aaa.aaa
PXE-E32: TFTP open timeout
PXE-M0F: Exiting Intel Boot Agent.

## Steps to Reproduce (for bugs)

In our case, because of the network infrastructure, the server is a bit slower to respond in that part of the world. 

## Context

Because the server takes longer to respond, TFTP's TTL expires and the session times out, ultimately failing the whole deployment process.

## Your Environment

Provisioner running CentOS 7, deploying to another bare-metal server.

gianarb:
Just sharing what I discovered. We use a lib called [betawaffle/tftp-go](https://github.com/betawaffle/tftp-go). It serves a UDP server with a handler that implements TFTP. From my understanding, the timeout can't be changed:

https://github.com/betawaffle/tftp-go/blob/master/server.go#L164

serve() does not accept a timeout and its value is set by default to `3s`. @mmlb can you confirm or deny? I feel a bit lost :)

mmlb:
all true. I've been wanting to either fork that repo and make some changes for a while now or look into an alternative. This may be the time to do so... probably the fork ;).

gianarb:
@mmlb apparently we have a fork now 

https://github.com/tinkerbell/tftp-go

I think @mikemrm did that, can we update that fork and fix this issue?

jacobweinstock:
iPXE binaries served via tftp now have a configurable timeout. please see: https://github.com/tinkerbell/boots/blob/5faab799ed5c1ceea501df7b3768effbbd7261ed/cmd/boots/main.go#L401

PR #229, opened 2021-12-19
Move to serving iPXE via github.com/tinkerbell/boots-ipxe:

jacobweinstock:
## Description

This is part of the Boots refactor proposal https://github.com/tinkerbell/proposals/tree/main/proposals/0022.
This isolates and simplifies serving iPXE binaries and allows iPXE binaries to be served from a remote location.

## Why is this needed

Partially Addresses: #66 (this PR would provide the same timeout for all machines, not a per machine timeout.)
Fixes: #213 
Fixes: #130 

## How Has This Been Tested?

Tested with the Tinkerbell Sandbox; Vagrant and Virtualbox quick start.

## Checklist:

I have:

- [ ] added unit or e2e tests


cmd/boots/dhcp.go
```diff
@@ -97,6 +104,8 @@ func (d dhcpHandler) serveDHCP(w dhcp4.ReplyWriter, req *dhcp4.Packet) {
 		return
 	}
 	span.End()
+	j.HttpServerFQDN = d.httpServerFQDN
```
rgl, line 107: nitpick: uppercase `j.HttpServerFQDN` to `j.HTTPServerFQDN`?

cmd/boots/main.go
```diff
@@ -96,14 +115,72 @@ func main() {
...
+	ipportHTTP, err := netaddr.ParseIPPort(c.HTTPAddr)
+	if err != nil {
+		mainlog.Fatal(err)
+	}
+	var nextServer net.IP
+	var httpServerFQDN string
+	if *rTFTP == "" && *rHTTP == "" {
+		mainlog.Info("serving ipxe binaries via tftp and http")
+		g.Go(func() error {
+			ipportTFTP, err := netaddr.ParseIPPort(c.TFTPAddr)
```
rgl, line 128: I would love to have a way to disable TFTP :-)

cmd/boots/dhcp.go
jacobweinstock: ah yes, thank you. update.

cmd/boots/main.go
jacobweinstock: just TFTP, right? still serve iPXE binaries via HTTP. 

Is this because your hardware doesn't use TFTP to PXE boot?
I know this isn't disabling TFTP, but is setting `tftp-addr` to `127.0.0.1:8888` (for example) a satisfactory workaround?

Regardless, I'm looking into what it looks like to incorporate this option.

rgl: > just TFTP, right? still serve iPXE binaries via HTTP.

Yes.

> Is this because your hardware doesn't use TFTP to PXE boot? I know this isn't disabling TFTP, but is setting the cli flag `-tftp-addr` to `127.0.0.1:8888` (for example) a satisfactory workaround?
> 
> Regardless, I'm looking into what it looks like to incorporate this option.

When its running in a container I guess it would.

Maybe just set it to an empty string? Or `--no-tftp`? Or `--no-tftp-addr`? I can even send a PR, we have to figure out which form of argument controls this.

jacobweinstock: `no-tftp`, `tftp-disabled`, `disable-tftp`, etc all seem reasonable. Also, FYI: https://github.com/tinkerbell/boots-ipxe/pull/20

rgl: @jacobweinstock, thx for https://github.com/tinkerbell/boots-ipxe/pull/20!

Adding the flag to the repository is the only thing missing? Do you plan on doing it? Or is it okay to wait for my PR (which I cannot say when I'll do it; maybe after a couple of weeks)?

jacobweinstock: I'm waiting for 2 more PRs to land in https://github.com/tinkerbell/ipxedust/pulls and then I'll be updating this PR. I can add the flag then, assuming it's before you've started any work :)

jacobweinstock: @rgl, PRs have landed. I've updated this draft PR. If you have a moment, please check it out and let me know what you think. thank you!

rgl:
I was able to successfully use this branch, but I have a couple of doubts/suggestions.

https://github.com/jacobweinstock/boots/commit/b3807550c1fd2bc97b386da10d3bcee03ce22bae seems to have dropped the port number from `-tftp-addr` argument and that broke the sandbox:

```
panic: ParseIP("10.3.0.2:69"): unexpected character (at ":69")
goroutine 1 [running]:
main.main()
	/root/tinkerbell-boots/cmd/boots/main.go:174 +0xbfe
```

Maybe boots can parse an address with the port and panic if its not 69? The reasoning being that all the other `addr` flags accept a port.

The boots flags now have an `ihttp` variants that are not clear to understand:

```console
# ./cmd/boots/boots -h
USAGE
  Run Boots server for provisioning

FLAGS
  -dhcp-addr 192.168.121.96:67     IP and port to listen on for DHCP.
  -http-addr 192.168.121.96:80     local IP and port to listen on for the serving iPXE files via HTTP.
  -ihttp-addr 0.0.0.0:8080         local IP and port to listen on for serving iPXE binaries via HTTP.
  -ihttp-disabled=false            disable serving iPXE binaries via HTTP
  -ihttp-timeout 5s                local iPXE HTTP server requests timeout
  -log-level info                  log level
  -remote-ihttp-addr ...           remote IP and port where iPXE binaries are served via HTTP.
  -remote-tftp-addr ...            remote IP where iPXE binaries are served via TFTP.
  -syslog-addr 192.168.121.96:514  IP and port to listen on for syslog messages.
  -tftp-addr 0.0.0.0               local IP to listen on for serving iPXE binaries via TFTP.
  -tftp-disabled=false             disable serving iPXE binaries via TFTP
  -tftp-timeout 5s                 local iPXE TFTP server requests timeout
```

Its not clear the difference between `-http-addr` and `-ihttp-addr` because their description seems to slightly different, but say more-or-less the same thing. Which makes it unclear why we need both and/or why they do not default to the same value.

Defaulting the `-ihttp-addr` to port `8080` conflicts with the sandbox configuration (that port is used by `docker-compose.yml` to expose `osie-bootloader` to the host), so I think it should be changed to something else.

Its not clear what is the meaning of `...` in the `-remote` flags.

jacobweinstock:
> I was able to successfully use this branch, but I have a couple of doubts/suggestions.
> 
> [jacobweinstock@b380755](https://github.com/jacobweinstock/boots/commit/b3807550c1fd2bc97b386da10d3bcee03ce22bae) seems to have dropped the port number from `-tftp-addr` argument and that broke the sandbox:
> 
> ```
> panic: ParseIP("10.3.0.2:69"): unexpected character (at ":69")
> goroutine 1 [running]:
> main.main()
> 	/root/tinkerbell-boots/cmd/boots/main.go:174 +0xbfe
> ```
> 
> Maybe boots can parse an address with the port and panic if its not 69? The reasoning being that all the other `addr` flags accept a port.
> 
> The boots flags now have an `ihttp` variants that are not clear to understand:
> 
> ```
> # ./cmd/boots/boots -h
> USAGE
>   Run Boots server for provisioning
> 
> FLAGS
>   -dhcp-addr 192.168.121.96:67     IP and port to listen on for DHCP.
>   -http-addr 192.168.121.96:80     local IP and port to listen on for the serving iPXE files via HTTP.
>   -ihttp-addr 0.0.0.0:8080         local IP and port to listen on for serving iPXE binaries via HTTP.
>   -ihttp-disabled=false            disable serving iPXE binaries via HTTP
>   -ihttp-timeout 5s                local iPXE HTTP server requests timeout
>   -log-level info                  log level
>   -remote-ihttp-addr ...           remote IP and port where iPXE binaries are served via HTTP.
>   -remote-tftp-addr ...            remote IP where iPXE binaries are served via TFTP.
>   -syslog-addr 192.168.121.96:514  IP and port to listen on for syslog messages.
>   -tftp-addr 0.0.0.0               local IP to listen on for serving iPXE binaries via TFTP.
>   -tftp-disabled=false             disable serving iPXE binaries via TFTP
>   -tftp-timeout 5s                 local iPXE TFTP server requests timeout
> ```
> 
> Its not clear the difference between `-http-addr` and `-ihttp-addr` because their description seems to slightly different, but say more-or-less the same thing. Which makes it unclear why we need both and/or why they do not default to the same value.
> 
> Defaulting the `-ihttp-addr` to port `8080` conflicts with the sandbox configuration (that port is used by `docker-compose.yml` to expose `osie-bootloader` to the host), so I think it should be changed to something else.
> 
> Its not clear what is the meaning of `...` in the `-remote` flags.

Great feedback! thank you @rgl for taking a look. I've updated the PR to hopefully address all your concerns. If you have some cycles please do check it out.

cmd/boots/main.go
```diff
@@ -97,13 +137,163 @@ func main() {
...
+		Log:                  defaultLogger(cfg.logLevel),
+		EnableTFTPSinglePort: true,
+		TFTP:                 ipxedust.ServerSpec{Disabled: true},
+		HTTP:                 ipxedust.ServerSpec{Disabled: true},
+	}
+
+	if cfg.remoteTFTPAddr == "" { // use local iPXE binary service for TFTP
+		if !cfg.iTFTPDisabled {
+			ip := cfg.ipxe.TFTPAddr
+			if strings.Contains(cfg.ipxe.TFTPAddr, ":") {
```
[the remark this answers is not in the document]
jacobweinstock: ah yes. updated.

cmd/boots/main.go
```diff
@@ -97,13 +137,163 @@ func main() {
...
+				ip = strings.Split(cfg.ipxe.TFTPAddr, ":")[0]
+				port := strings.Split(cfg.ipxe.TFTPAddr, ":")[1]
+				mainlog.With("providedPort", port).Info("warning: only port 69 is supported for TFTP")
```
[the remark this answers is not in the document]
jacobweinstock: yes, updated.

rgl:
I had a chance to try this branch in KVM VMs.

I was able to TFTP boot BIOS and UEFI KVM VMs work.

Disabling tinkerbell-boots TFTP (with `-tftp-disabled=true`) I was able to HTTP boot BIOS KVM VMs.

But failed to HTTP boot UEFI KVM VMs. Boots replies with `301` when the UEFI firmware tried to download `http://10.3.0.2:80/ipxe//ipxe.efi`; it does that because of that extra forward slash in the uri/filename. Here's the same request using httpie:

```console
# http http://10.3.0.2/ipxe//ipxe.efi
HTTP/1.1 301 Moved Permanently
Content-Length: 49
Content-Type: text/html; charset=utf-8
Date: Sun, 09 Jan 2022 20:44:08 GMT
Location: /ipxe/ipxe.efi

<a href="/ipxe/ipxe.efi">Moved Permanently</a>.
```

That 301 seems to fail the UEFI firmware HTTP boot.

Here are the boots logs:

```json
{"level":"info","ts":1641759718.1115346,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:105","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"recv","mac":"08:00:27:00:00:02","via":"0.0.0.0","iface":"eth1","xid":"\"5c:c6:5a:ca\"","type":"DHCPREQUEST"}
{"level":"info","ts":1641759718.1117184,"caller":"boots/dhcp.go:80","msg":"parsed option82/circuitid","service":"github.com/tinkerbell/boots","pkg":"main","mac":"08:00:27:00:00:02","circuitID":""}
{"level":"info","ts":1641759718.115889,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:61","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"send","mac":"08:00:27:00:00:02","dst":"255.255.255.255","iface":"eth1","xid":"\"5c:c6:5a:ca\"","type":"DHCPACK","address":"10.3.0.12","next_server":"10.3.0.2","filename":"ipxe.efi"}
{"level":"info","ts":1641759718.1210024,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:105","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"recv","mac":"08:00:27:00:00:02","via":"0.0.0.0","iface":"eth1","xid":"\"5c:c6:5a:cb\"","type":"DHCPDISCOVER"}
{"level":"info","ts":1641759718.1211138,"caller":"boots/dhcp.go:80","msg":"parsed option82/circuitid","service":"github.com/tinkerbell/boots","pkg":"main","mac":"08:00:27:00:00:02","circuitID":""}
{"level":"info","ts":1641759718.1235836,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:61","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"send","mac":"08:00:27:00:00:02","dst":"255.255.255.255","iface":"eth1","xid":"\"5c:c6:5a:cb\"","type":"DHCPOFFER","address":"10.3.0.12","next_server":"10.3.0.2","filename":"http://10.3.0.2:80/ipxe//ipxe.efi"}
{"level":"info","ts":1641759722.111614,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:105","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"recv","mac":"08:00:27:00:00:02","via":"0.0.0.0","iface":"eth1","xid":"\"5c:c6:5a:cb\"","type":"DHCPREQUEST"}
{"level":"info","ts":1641759722.1117978,"caller":"boots/dhcp.go:80","msg":"parsed option82/circuitid","service":"github.com/tinkerbell/boots","pkg":"main","mac":"08:00:27:00:00:02","circuitID":""}
{"level":"info","ts":1641759722.1161463,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:61","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"send","mac":"08:00:27:00:00:02","dst":"255.255.255.255","iface":"eth1","xid":"\"5c:c6:5a:cb\"","type":"DHCPACK","address":"10.3.0.12","next_server":"10.3.0.2","filename":"http://10.3.0.2:80/ipxe//ipxe.efi"}
{"level":"info","ts":1641759722.1183524,"caller":"httplog/httplog.go:37","msg":"","service":"github.com/tinkerbell/boots","pkg":"http","event":"ss","method":"HEAD","uri":"/ipxe//ipxe.efi","client":"10.3.0.12","duration":0.00008226,"status":301}
{"level":"info","ts":1641759722.1233978,"caller":"httplog/httplog.go:37","msg":"","service":"github.com/tinkerbell/boots","pkg":"http","event":"ss","method":"GET","uri":"/ipxe//ipxe.efi","client":"10.3.0.12","duration":0.000081108,"status":301}
```

jacobweinstock:
> I had a chance to try this branch in KVM VMs.
> 
> I was able to TFTP boot BIOS and UEFI KVM VMs work.
> 
> Disabling tinkerbell-boots TFTP (with `-tftp-disabled=true`) I was able to HTTP boot BIOS KVM VMs.
> 
> But failed to HTTP boot UEFI KVM VMs. Boots replies with `301` when the UEFI firmware tried to download `http://10.3.0.2:80/ipxe//ipxe.efi`; it does that because of that extra forward slash in the uri/filename. Here's the same request using httpie:
> 
> ```
> # http http://10.3.0.2/ipxe//ipxe.efi
> HTTP/1.1 301 Moved Permanently
> Content-Length: 49
> Content-Type: text/html; charset=utf-8
> Date: Sun, 09 Jan 2022 20:44:08 GMT
> Location: /ipxe/ipxe.efi
> 
> <a href="/ipxe/ipxe.efi">Moved Permanently</a>.
> ```
> 
> That 301 seems to fail the UEFI firmware HTTP boot.
> 
> Here are the boots logs:
> 
> ```json
> {"level":"info","ts":1641759718.1115346,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:105","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"recv","mac":"08:00:27:00:00:02","via":"0.0.0.0","iface":"eth1","xid":"\"5c:c6:5a:ca\"","type":"DHCPREQUEST"}
> {"level":"info","ts":1641759718.1117184,"caller":"boots/dhcp.go:80","msg":"parsed option82/circuitid","service":"github.com/tinkerbell/boots","pkg":"main","mac":"08:00:27:00:00:02","circuitID":""}
> {"level":"info","ts":1641759718.115889,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:61","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"send","mac":"08:00:27:00:00:02","dst":"255.255.255.255","iface":"eth1","xid":"\"5c:c6:5a:ca\"","type":"DHCPACK","address":"10.3.0.12","next_server":"10.3.0.2","filename":"ipxe.efi"}
> {"level":"info","ts":1641759718.1210024,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:105","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"recv","mac":"08:00:27:00:00:02","via":"0.0.0.0","iface":"eth1","xid":"\"5c:c6:5a:cb\"","type":"DHCPDISCOVER"}
> {"level":"info","ts":1641759718.1211138,"caller":"boots/dhcp.go:80","msg":"parsed option82/circuitid","service":"github.com/tinkerbell/boots","pkg":"main","mac":"08:00:27:00:00:02","circuitID":""}
> {"level":"info","ts":1641759718.1235836,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:61","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"send","mac":"08:00:27:00:00:02","dst":"255.255.255.255","iface":"eth1","xid":"\"5c:c6:5a:cb\"","type":"DHCPOFFER","address":"10.3.0.12","next_server":"10.3.0.2","filename":"http://10.3.0.2:80/ipxe//ipxe.efi"}
> {"level":"info","ts":1641759722.111614,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:105","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"recv","mac":"08:00:27:00:00:02","via":"0.0.0.0","iface":"eth1","xid":"\"5c:c6:5a:cb\"","type":"DHCPREQUEST"}
> {"level":"info","ts":1641759722.1117978,"caller":"boots/dhcp.go:80","msg":"parsed option82/circuitid","service":"github.com/tinkerbell/boots","pkg":"main","mac":"08:00:27:00:00:02","circuitID":""}
> {"level":"info","ts":1641759722.1161463,"caller":"dhcp4-go@v0.0.0-20190402165401-39c137f31ad3/handler.go:61","msg":"","service":"github.com/tinkerbell/boots","pkg":"dhcp","pkg":"dhcp","event":"send","mac":"08:00:27:00:00:02","dst":"255.255.255.255","iface":"eth1","xid":"\"5c:c6:5a:cb\"","type":"DHCPACK","address":"10.3.0.12","next_server":"10.3.0.2","filename":"http://10.3.0.2:80/ipxe//ipxe.efi"}
> {"level":"info","ts":1641759722.1183524,"caller":"httplog/httplog.go:37","msg":"","service":"github.com/tinkerbell/boots","pkg":"http","event":"ss","method":"HEAD","uri":"/ipxe//ipxe.efi","client":"10.3.0.12","duration":0.00008226,"status":301}
> {"level":"info","ts":1641759722.1233978,"caller":"httplog/httplog.go:37","msg":"","service":"github.com/tinkerbell/boots","pkg":"http","event":"ss","method":"GET","uri":"/ipxe//ipxe.efi","client":"10.3.0.12","duration":0.000081108,"status":301}
> ```

Thanks for testing. good catch. just pushed a change that should resolve this issue. I will try to test it too. thanks again for all the collaboration on this!

rgl:
@jacobweinstock, it did work! the VM can now UEFI HTTP boot :-)

there are couple of improvement that we should do:

1. when TFTP is disabled, tinkerbell-boots should not send a next server ip address nor a boot file name.
2. support the HEAD method (currently, the HTTP boot client sends a HEAD request, but tinkerbell-boots fails it with 405 Method Not Allowed) and return the Content-Length header. these are required for booting the Seeed Studio Odyssey X86J4105 board ([as mentioned in my original HTTP Boot PR](https://github.com/tinkerbell/boots/pull/216/files#diff-a8783392d81a537e4fb749c1b3e44b23ad376a7ee72440e0cefe1597ddd28acfR24-R27)).

rgl:
@jacobweinstock, using the latest commits, I can now UEFI HTTP boot my Seeed Studio Odyssey X86J4105 :-)

Just pending your thoughts about:

1. when TFTP is disabled, tinkerbell-boots should not send a next server ip address nor a boot file name.

jacobweinstock:
> @jacobweinstock, using the latest commits, I can now UEFI HTTP boot my Seeed Studio Odyssey X86J4105 :-)
> 
> Just pending your thoughts about:
> 
> 1. when TFTP is disabled, tinkerbell-boots should not send a next server ip address nor a boot file name.

nice! glad that is working. I think 1. is a good idea. I'm looking into/working on that now.

jacobweinstock:
Hey @rgl, I think there is a behavior here that might be on purpose. My testing is showing that when I stop sending the next server and filename when TFTP is disabled then if/when a TFTP (PXE) client requests to network boot the client will have to wait for the network boot to timeout. If TFTP is disabled but we still send the next server and filename the client will try to TFTP but fail. This looks to be a much much shorter timeout than waiting for the network boot to timeout. This behavior could be desired for some clients. 

I'm still thinking about this, but curious about your thoughts.

rgl:
I see, I'm OK will leaving it as-is. I guess that behavior would also be useful with clients that do not let the user disable TFTP.

job/dhcp.go
```diff
@@ -139,15 +139,15 @@ func (j Job) setPXEFilename(rep *dhcp4.Packet, isPacket, isARM, isUEFI, isHTTPCl
...
 
 	var filename string
-	if !isPacket {
+	if !isTinkerbellIPXE {
 		if j.PArch() == "hua" || j.PArch() == "2a2" {
-			filename = "ipxe/snp-hua.efi"
+			filename = "snp.efi"
 		} else if isARM {
-			filename = "ipxe/snp-nolacp.efi"
+			filename = "snp.efi"
```
mmlb, line 146: This caught my eye. Have you been able to test this against the hua/2a2 machines in our fleet? The snp-hua.efi was definitely a hard requirement as no other ipxe build would boot. If this is no longer the case then we can drop the `PAarch` check and should be called out in the commit message explicitly.

cmd/boots/main.go
```diff
@@ -97,13 +138,163 @@ func main() {
...
+			if strings.Contains(ip, ":") {
+				ip = strings.Split(ip, ":")[0]
+				if port := strings.Split(cfg.ipxe.TFTPAddr, ":")[1]; port != "69" {
+					mainlog.With("providedPort", port).Info("warning: only port 69 is supported for TFTP")
+				}
+			}
+			ipportTFTP, err := netaddr.ParseIPPort(ip + ":69")
+			if err != nil {
+				mainlog.Fatal(err)
+			}
```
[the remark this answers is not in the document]
mmlb: follow up: I went looking through git history and saw the commit that made it a warning but there's no explanation for it. Please add more info to the commit message If you stick with the warning.

rules.mk
```diff
@@ -37,7 +37,7 @@ export PATH
 endif
 
 # parses tools.go and returns the tool name prefixed with bin/
-toolsBins := $(addprefix bin/,$(notdir $(shell awk -F'"' '/^\s*_/ {print $$2}' tools.go)))
+toolsBins := $(addprefix bin/,$(notdir $(shell awk -F'"' '/^\s*/ {print $$2}' tools.go)))
```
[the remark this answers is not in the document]
mmlb: follow up: I checked the commit that did this but didn't see the underlying issue. If we have to keep the `_` out it would be good to have the cause in the message.

ScottGarman (approved):

ScottGarman (changes requested):
I hit the approval too quickly and now see Manny had some worthwhile suggestions.

cmd/boots/main.go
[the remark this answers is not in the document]
jacobweinstock: https://github.com/tinkerbell/boots/pull/229/commits/b3807550c1fd2bc97b386da10d3bcee03ce22bae

_"github.com/packethost/dhcp4-go does not allow for specifying
a port for the TFTP server. This removes cli flags indicating
that a port should be specified. This also means that local or
remote TFTP servers must be listening on port 69 in order to
interoperate with Boots."_

This PR started out dropping support for a port in the local TFTP address, but the port was [requested](https://github.com/tinkerbell/boots/pull/229#issuecomment-1008262077) back in by @rgl. Because of the git comment above the port must be 69 and so we statically set it regardless of what the user provides. So the warning is just that. "only port 69 is supported for TFTP".

rules.mk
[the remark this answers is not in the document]
jacobweinstock: https://github.com/tinkerbell/boots/pull/229/commits/c407c9e5277ba73cefdb956d678998ddd84ab6ea

_"Get make gen working on MacOS"_

This change doesn't seem to be working anymore. I just tested on another Mac. I will need to revisit this.

job/dhcp.go
jacobweinstock: I have not tested against any hua/2a2. Any idea how to build this iPXE binary? it should/needs to be built in the open like the others in github.com/tinkerbell/ipxedust but seems it's just been saved in [git-lfs](https://github.com/tinkerbell/boots/blob/ba3a3fef424ebfd7125b08ae99dcb9631bc911a8/.gitattributes).

mmlb: I haven't tried building this recently (it's been at least 2 years) but every time I tried it failed to boot. As far as I know its a magic golden image, one of a kind. That being said it *would* be good to see if recent ipxe builds will boot correctly on them.

cmd/boots/main.go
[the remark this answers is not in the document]
mmlb: In that message he also mentions that maybe we should panic if a different port was given. I don't think ignoring the user is the correct move here. We should either exit if an invalid port is given or go back to only taking an ip address and updating sandbox so it works there. The current setup just seems to be the most confusing route.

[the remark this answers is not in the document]
jacobweinstock: Hey @rgl, curious about your thoughts here. Allowing a user the ability to specify a port feels kind of misleading. Like saying, "you can specify any port as long as it's port 69". Thoughts on only taking an IP for TFTP and then updating the sandbox? I do feel that since we can't allow customization of the port yet, we shouldn't allow the user to specify one.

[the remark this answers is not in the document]
rgl: My only concern is about the cli argument names and their values; all but the tftp one receive a port:

```
  -dhcp-addr 192.168.121.96:67     IP and port to listen on for DHCP.
  -http-addr 192.168.121.96:80     local IP and port to listen on for the serving iPXE files via HTTP.
  -syslog-addr 192.168.121.96:514  IP and port to listen on for syslog messages.
  -tftp-addr 0.0.0.0               local IP to listen on for serving iPXE binaries via TFTP.
```

the `-addr` suffix seems to infer the same type/syntax can be used in all of them.

I'm fine in boots crashing when the user does not set a port to 69 (omitting it should be fine too; it would default to 69). 

Having a port seems more consistent and explicit.

But at the end of the day, that is just my opinion. I will use it anyways :-)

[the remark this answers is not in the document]
jacobweinstock: @mmlb, @rgl  pushed changes to this. Thoughts?

rules.mk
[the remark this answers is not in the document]
jacobweinstock: from testing this. the `awk` line with the `_` isn't working for me at all. tested in a Linux container and on a mac. once the `_` is removed, then it works on both.

cmd/boots/main.go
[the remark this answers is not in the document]
rgl: @jacobweinstock, looking good from here :-)

mmlb (changes requested):

rules.mk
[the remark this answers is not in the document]
jacobweinstock: yeah, works here. thanks. i've updated the PR with this.

job/dhcp.go
jacobweinstock: Hey @mmlb, I tested this against Equinix Metal machines with plan type c1.large.arm.xda ([boots ref](https://github.com/tinkerbell/boots/blob/aefc60d0f405bd0fc0c56fc000f7b302801171ff/job/helpers.go#L47)) and against plan type Type HUA (baremetal_hua) ([boots ref](https://github.com/tinkerbell/boots/blob/aefc60d0f405bd0fc0c56fc000f7b302801171ff/job/helpers.go#L53)) and they were both successful.

mmlb: *awesome*!

jacobweinstock:
Hey @mmlb and @ScottGarman, when you have some cycles, mind checking this one out?

README.md
```diff
@@ -15,35 +15,38 @@ As boots runs a DHCP server, it's often asked if it is safe to run without any n
...
+USAGE
+  Run Boots server for provisioning
+
+FLAGS
+  -dhcp-addr          IP and port to listen on for DHCP. (default "0.0.0.0:67")
+  -http-addr          local IP and port to listen on for the serving iPXE binaries and files via HTTP. (default "0.0.0.0:80")
+  -ihttp-disabled     disable serving iPXE binaries via HTTP. (default "false")
+  -log-level          log level. (default "info")
+  -remote-ihttp-addr  remote IP and port where iPXE binaries are served via HTTP. Overrides -http-addr for iPXE binaries only.
+  -remote-tftp-addr   remote IP where iPXE binaries are served via TFTP. Overrides -tftp-addr.
```
mmlb, line 42: can we use clearer flags here? Maybe something like:
```
-ipxe-enable-http
-ipxe-enable-tftp
-ipxe-remote-http-addr
-ipxe-remote-tftp-addr
```
I think this is nicer because it groups the the ipxe stuff together which extends to the enabling serving via http and/or tftp. Also flips from double negative to straight positive which I think is also easier to reason about. Also avoids having to think about what `ihttp` could mean.

cmd/boots/main.go
```diff
@@ -17,16 +30,19 @@ import (
 	"github.com/tinkerbell/boots/metrics"
 	"github.com/tinkerbell/boots/packet"
 	"github.com/tinkerbell/boots/syslog"
-	"github.com/tinkerbell/boots/tftp"
+	"github.com/tinkerbell/ipxedust"
+	"github.com/tinkerbell/ipxedust/ihttp"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"golang.org/x/sync/errgroup"
+	"inet.af/netaddr"
```
mmlb, line 39: can you remove this empty line here? (hmm boots needs some linter updates)

cmd/boots/main.go
```diff
@@ -97,13 +140,162 @@ func main() {
...
+	lg = lg.WithName("github.com/tinkerbell/ipxedust")
+	ipxe := &ipxedust.Server{
+		Log:                  lg,
+		EnableTFTPSinglePort: true,
+		TFTP:                 ipxedust.ServerSpec{Disabled: true},
+		HTTP:                 ipxedust.ServerSpec{Disabled: true},
+	}
+
+	if cfg.remoteTFTPAddr == "" { // use local iPXE binary service for TFTP
+		if !cfg.iTFTPDisabled {
```
mmlb, line 157: imo this would also be easier to read as:
```go
if cfg.iPXETFTPEnabled {
...
```

cmd/boots/main.go
```diff
@@ -97,13 +140,162 @@ func main() {
 		}
 	}()
 
-	mainlog.Info("serving tftp")
-	go ServeTFTP()
-	mainlog.Info("serving dhcp")
-	go ServeDHCP()
+	var nextServer net.IP
+	var httpServerFQDN string
```
mmlb, line 144: small nit: it would be nice if these were define close to when they were actually used. So nextServer right before the `if cfg.RemoteTFTPAddr` check and httpServerFQDN right before the `if cfg.remoteiHTTPAddr` with the other related vars.

mmlb (changes requested):

cmd/boots/main.go
jacobweinstock: update

cmd/boots/main.go
jacobweinstock: updated

cmd/boots/main.go
jacobweinstock: updated

README.md
jacobweinstock: updated

ScottGarman (approved):

mmlb (approved):

job/dhcp.go
```diff
@@ -174,5 +175,5 @@ func (j Job) setPXEFilename(rep *dhcp4.Packet, isPacket, isARM, isUEFI, isHTTPCl
 		return
 	}
 
-	dhcp.SetFilename(rep, filename, conf.PublicIPv4, isHTTPClient, conf.PublicFQDN)
+	dhcp.SetFilename(rep, filename, j.NextServer, isHTTPClient, j.HTTPServerFQDN)
```
nshalman, line 178: This might be the root cause that led to the bug we are trying to fix in #246

job/dhcp.go
```diff
@@ -141,13 +140,13 @@ func (j Job) setPXEFilename(rep *dhcp4.Packet, isPacket, isARM, isUEFI, isHTTPCl
...
+			filename = "snp.efi"
 		} else if isARM {
-			filename = "ipxe/snp-nolacp.efi"
+			filename = "snp.efi"
 		} else if isUEFI {
-			filename = "ipxe/ipxe.efi"
+			filename = "ipxe.efi"
 		} else {
-			filename = "ipxe/undionly.kpxe"
+			filename = "undionly.kpxe"
```
ScottGarman, line 150: Was there a reason you removed the `/ipxe/` prefix from these files? I'm running into issues trying to deploy this in production where we're getting a 404 error when these files are requested from the root path. My understanding of the way things used to be is that auto.ipxe was served from the root, and everything else was from the `/ipxe/` subdirectory. I'm looking to possibly revert this behavior in a new PR but don't fully understand the context for the change.

jacobweinstock: This prefix, `ipxe/` was removed here but added back in, for HTTP netboot clients only, here: https://github.com/tinkerbell/boots/blob/fdd7a4fb385659328bb8c3c6625741936c00aaa5/job/dhcp.go#L178

@nshalman changed this behavior with his last commit. `j.HTTPServerFQDN` to `conf.PublicFQDN`. This feels like why you are seeing issues.

jacobweinstock: @ScottGarman and @nshalman, happy to jump on a call tomorrow if that would be helpful.

PR #236, opened 2022-02-23
Change from "packet" to "metal"

edw-eqix:
## Description

Change ipxe_cloud_config from "packet" to "metal" to match the change made in https://github.com/netbootxyz/netboot.xyz/pull/1076.

## Why is this needed

Fixes: #

## How Has This Been Tested?

`make test` and will be tested in one production facility to see that netbooting works without applying a work-around via the IDRAC console.

Fixes a problem where c3.medium Dell hardware does not configure the serial console correctly on netboot.

## Checklist:

I have:

- [ ] added unit or e2e tests


mmlb:
oh hmm this is a *little* worrisome actually. This can be seen as a breaking behavior. I'm not sure if any thing else but netboot uses ipxe_cloud_config but I'd like to not break them.

chrisdoherty4:
Closing as it seems stale.

PR #238, opened 2022-03-14
http: Exit if "xff.New" fails.

stephen-fox:
## Description
If the call to "xff.New" failed, it would return a nil
"XFF" pointer. This would not result in a nil pointer
deref, and the program would continue on as normal.

While there is (currently) no code path that allows this
to happen, it feels reasonable to guard against it just
in case a new code path emerges.

## How Has This Been Tested?
The modified code has no unit tests... I can spend
some time adding some if desired.

As far as I can tell, there is no way to hit this code
path. This might change in the future though, hence
this code change.

## Checklist:

I have:

- [ ] added unit or e2e tests


jacobweinstock (approved):
seems reasonable to me. thanks!

PR #240, opened 2022-03-18
OSIE installer changes: Eclypsium removal and Vendor Services URL addition

ScottGarman:
This removes passing the Eclypsium auth token passthrough in boots, and adds a new env-based variable which is used to customize the url used to download OS artifacts from.


markyjackson-taulia (approved):
/lgtm

installers/osie/main.go
```diff
@@ -79,6 +79,9 @@ func kernelParams(ctx context.Context, action, state string, j job.Job, s ipxe.S
 	s.Args("parch=${parch}")
 	s.Args("packet_action=${action}")
 	s.Args("packet_state=${state}")
+	if conf.VendorServicesURL != "" {
+		s.Args("osie_vendors_url=" + conf.VendorServicesURL)
```
rgl, line 83: I'm not really sure why we need this osie thing. But probably the variable should be called `conf.OsieVendorsUrl` (and the environment variable should also have a name like this) instead? That would make things more explicit by sharing the same name.

ScottGarman: An idea came up recently to try to move all of the kernel cmdline options into env vars that are composed outside of the boots code base, and I'm hopeful that this will allow us to move a lot of the osie-specifc code from boots. In the meantime, I've taken your advice and renamed to the conf and env var to use osie as the prefix.

nshalman (approved):

ScottGarman:
Was having weird issues trying to rebase this PR today, likely related to some GitHub outage issues. I re-created it as https://github.com/tinkerbell/boots/pull/242 and completed the work there.

PR #242, opened 2022-03-22
OSIE installer changes: Eclypsium removal and Vendor Services URL addition

ScottGarman:
This removes passing the Eclypsium auth token passthrough in boots, and adds a new env-based variable which is used to customize the url used to download OS artifacts from.


nshalman (approved):

nshalman:
Fixes #240

PR #246, opened 2022-03-28
Always set HTTPServerFQDN from conf.PublicFQDN

nshalman:
## Description

We saw an issue where DHCP responses from boots contained:
```http://:80/ipxe/undionly.kpxe```

We think this should fix that.

Suggestions on how a test case could have caught this are very much welcome!!

## Why is this needed

This is a regression when running in production.

## How Has This Been Tested?
 - [ ] In a Sandbox

## Checklist:

I have:

- [ ] added unit or e2e tests

### CODE REVIEW

**Comment by nshalman** [comment]
@jacobweinstock if you have a moment to take a look too, I would greatly appreciate it!

**Review by ScottGarman — APPROVED** [review]

File: job/dhcp.go [hunk 1]
```diff
@@ -175,5 +174,10 @@ func (j Job) setPXEFilename(rep *dhcp4.Packet, isTinkerbellIPXE, isARM, isUEFI,
 		return
 	}
 
+	// If the client speaks HTTP, use the PublicFQDN.
+	if isHTTPClient {
```
jacobweinstock, line 178: hmmm....this might not need to be wrapped in an `if` statement at all. line 182 (`dhcp.SetFilename`) does this same check. https://github.com/tinkerbell/boots/blob/fdd7a4fb385659328bb8c3c6625741936c00aaa5/dhcp/pxe.go#L143

jacobweinstock: one option could be to just use the previous code, like you referenced in [this](https://github.com/tinkerbell/boots/pull/229#discussion_r836704325) comment.

nshalman: At that rate, at least for now, I could just modify the function call directly:
```diff
- dhcp.SetFilename(rep, filename, j.NextServer, isHTTPClient, j.HTTPServerFQDN)
+ dhcp.SetFilename(rep, filename, j.NextServer, isHTTPClient, conf.PublicFQDN)
```

nshalman: Changed.

nshalman: I imagine there was a purpose behind that other change which in the future we should untangle and move forward with at some point, but for now I think this will fix our production issue.

ScottGarman (approved):

jacobweinstock (approved):

nshalman:
Bad news. This doesn't quite solve the problem.

PR #247, opened 2022-03-28
WIP: Serve non-auto.ipxe files from /ipxe/ when using http

ScottGarman:
This fixes a recent regression which caused files such as `undionly,kpxe` to be served from the root url path of boots. Only auto.ipxe should be served from the root, everything else comes from /ipxe/.

### DIFF
--- a/dhcp/pxe.go
+++ b/dhcp/pxe.go
@@ -141,7 +141,13 @@ func SetFilename(rep *dhcp4.Packet, filename string, nextServer net.IP, httpClie
 	rep.SetSIAddr(nextServer.To4()) // next-server: IP address of the TFTP/HTTP Server.
 
 	if httpClient {
-		urlPath := path.Join(httpServerFQDN, filename)
+		urlPath := ""
+		if filename == "auto.ipxe" {
+			urlPath = path.Join(httpServerFQDN, filename)
+		} else {
+			// we serve everything other than auto.ipxe from /ipxe/
+			urlPath = path.Join(httpServerFQDN, "ipxe", filename)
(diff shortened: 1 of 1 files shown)


jacobweinstock:
Hey. Just FYI, iPXE binaries served via HTTP are only available from `/ipxe/`. For example: `/ipxe/undionly.kpxe`. See the following references below.
https://github.com/tinkerbell/boots/blob/7df4b4eb341b184a76e0f0b3b8a6d93114ceeed5/cmd/boots/main.go#L184
https://github.com/tinkerbell/boots/blob/7df4b4eb341b184a76e0f0b3b8a6d93114ceeed5/cmd/boots/http.go#L69

If an iPXE binary is requested from the root, like `/undionly.kpxe` for example, Boots will return a 200 but the file will be empty. This is because of this function: https://github.com/tinkerbell/boots/blob/7df4b4eb341b184a76e0f0b3b8a6d93114ceeed5/job/http.go#L15
It might be worth add something like this: `w.WriteHeader(http.StatusNotFound)` to the end of this `ServeFile` function, so that consumers will get a 404 instead of a 200.

nshalman:
> But with @nshalman last commit, the argument to [SetFilename](https://github.com/tinkerbell/boots/blob/7df4b4eb341b184a76e0f0b3b8a6d93114ceeed5/job/dhcp.go#L177) was changed to not use this value.

My commit doesn't fix the problem and is a good candidate for reverting / undoing alongside fixing this issue.

ScottGarman:
Closing this in favor of #249

PR #248, opened 2022-03-29
Make local registry and nginx optional

ptrivedi:
* Add environment variables for
	- OSIE path override
        - Custom URL for registry CA certificate
        - tink-worker image:tag and path
	- Registry certificate required
	- Template action image absolute path URIs
  These environment variables allow
	- Bypassing the requirement for OSIE artifacts to be under
          predefined paths
	- Specifying custom URL for downloading CA certificate for
          the registry, which can be passed on to Hook for tink-docker
	  setup
	- Specifying custom tink-worker image:tag and path, which can
	  be passed to hook for tink-worker image pull
	- Specifying whether a registry certificate is required and needs
	  to be downloaded from either the default path hosted under
	  workflow directory on MIRROR_HOST webserver or from one specified
	  via custom path mentioned above
	- Specifying whether to consider the action image URIs specified in
	  the template as absolute paths and skip prepending registry to
          these
* Make registry username/password optional to allow for public registry
  pulls

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/installers/osie/main.go
+++ b/installers/osie/main.go
@@ -99,14 +99,30 @@ func kernelParams(ctx context.Context, action, state string, j job.Job, s ipxe.S
 
 	if j.CanWorkflow() {
 		buildWorkerParams()
-		s.Args("docker_registry=" + dockerRegistry)
+		if len(dockerRegistry) > 0 {
+			s.Args("docker_registry=" + dockerRegistry)
+		}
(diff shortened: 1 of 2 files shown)


cprivite:
@ptrivedi Can you sign the DCO?

cprivite:
@dependabot reopen

PR #249, opened 2022-03-29
Disambiguate ipxe HTTP URLs from other boots URLs

nshalman:
## Description

With the splitting out of ipxedust, the goal is for boots to be able to send DHCP responses
that point to a separate ipxedust server. The code didn't fully make that distinction
all the way down to the DHCP responses.

## Why is this needed

Reverts #246 
Fixes the issue identified in #246

## How Has This Been Tested?

- [ ] In a sandbox
- [x] In EM Production

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/boots/dhcp.go
+++ b/cmd/boots/dhcp.go
@@ -26,10 +26,10 @@ import (
 func ServeDHCP(addr string, nextServer net.IP, ipxeBaseURL string, bootsBaseURL string) {
 	poolSize := env.Int("BOOTS_DHCP_WORKERS", runtime.GOMAXPROCS(0)/2)
 	handler := dhcpHandler{
-		pool:           workerpool.New(poolSize),
-		nextServer:     nextServer,
-		ipxeBaseURL:    ipxeBaseURL,
-		bootsBaseURL:   bootsBaseURL,
+		pool:         workerpool.New(poolSize),
+		nextServer:   nextServer,
+		ipxeBaseURL:  ipxeBaseURL,
+		bootsBaseURL: bootsBaseURL,
 	}
 	defer handler.pool.Stop()
 
@@ -44,10 +44,10 @@ func ServeDHCP(addr string, nextServer net.IP, ipxeBaseURL string, bootsBaseURL 
 }
 
 type dhcpHandler struct {
-	pool           *workerpool.WorkerPool
-	nextServer     net.IP
-	ipxeBaseURL    string
-	bootsBaseURL   string
+	pool         *workerpool.WorkerPool
+	nextServer   net.IP
+	ipxeBaseURL  string
+	bootsBaseURL string
 }
 
 func (d dhcpHandler) ServeDHCP(w dhcp4.ReplyWriter, req *dhcp4.Packet) {
--- a/cmd/boots/dhcp.go
+++ b/cmd/boots/dhcp.go
@@ -23,12 +23,13 @@ import (
 // ServeDHCP starts the DHCP server.
 // It takes the next server address (nextServer) for serving iPXE binaries via TFTP
 // and an IP:Port (httpServerFQDN) for serving iPXE binaries via HTTP.
-func ServeDHCP(addr string, nextServer net.IP, httpServerFQDN string) {
+func ServeDHCP(addr string, nextServer net.IP, ipxeBaseURL string, bootsBaseURL string) {
 	poolSize := env.Int("BOOTS_DHCP_WORKERS", runtime.GOMAXPROCS(0)/2)
 	handler := dhcpHandler{
 		pool:           workerpool.New(poolSize),
 		nextServer:     nextServer,
-		httpServerFQDN: httpServerFQDN,
+		ipxeBaseURL:    ipxeBaseURL,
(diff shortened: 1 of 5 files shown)


nshalman:
CC @jacobweinstock 

Also if folks have suggestions for better names for those variables, please let me know.

stephen-fox (changes requested):

cmd/boots/main.go
```diff
@@ -176,24 +176,25 @@ func main() {
 
 	var ipxeHandler func(http.ResponseWriter, *http.Request)
 	var ipxePattern string
-	var httpServerFQDN string
+	var ipxeBaseURL string
+	bootsBaseURL := conf.PublicFQDN
```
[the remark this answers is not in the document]
nshalman: That might actually make things a lot clearer. Let me take a look.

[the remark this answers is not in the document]
nshalman: Done.

stephen-fox (approved):
For variable names, what do you think about removing the word `base` from them?

Like: `ipxeBinsURL` (honestly, I cannot recall the other variable now).

Otherwise - looks good to me.

ScottGarman:
Limited production testing in EM infra has been going well this afternoon. :100:

ScottGarman (approved):

job/dhcp.go
```diff
@@ -139,7 +139,9 @@ func (j Job) setPXEFilename(rep *dhcp4.Packet, isTinkerbellIPXE, isARM, isUEFI,
 	}
 
 	var filename string
+	httpPrefix := conf.PublicFQDN
```
jacobweinstock, line 142: This doesn't align with Boots' design philosophy, "_Only func main has the right to decide which flags, env variables, config files are available to the user_" -- [ref](https://github.com/tinkerbell/boots/blob/main/docs/DESIGNPHILOSOPHY.md#:~:text=Only%20func%20main%20has%20the%20right%20to%20decide%20which%20flags%2C%20env%20variables%2C%20config%20files%20are%20available%20to%20the%20user)
Would you be open to passing another value in along with `ipxeBaseURL`?

jacobweinstock (changes requested):

job/dhcp.go
nshalman: So revert https://github.com/tinkerbell/boots/pull/249/commits/199e821d8c7658eb81b9c61f08fb73b655345562 ?

cmd/boots/main.go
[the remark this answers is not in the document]
nshalman: https://github.com/tinkerbell/boots/pull/249#discussion_r838044182 might be why we do it.

nshalman:
I think I'm going to revert https://github.com/tinkerbell/boots/pull/249/commits/199e821d8c7658eb81b9c61f08fb73b655345562 and we should finalize what the variable names should be and then get this merged.

cmd/boots/main.go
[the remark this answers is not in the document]
nshalman: I agree that it's clunky, but I think we should save that refactor for elsewhere, per @jacobweinstock's comment.

ScottGarman (approved):

job/dhcp.go
nshalman: I removed that commit, please take another look.

jacobweinstock: looks good, thank you!

cmd/boots/dhcp.go
```diff
@@ -23,12 +23,13 @@ import (
...
+func ServeDHCP(addr string, nextServer net.IP, ipxeBaseURL string, bootsBaseURL string) {
 	poolSize := env.Int("BOOTS_DHCP_WORKERS", runtime.GOMAXPROCS(0)/2)
 	handler := dhcpHandler{
-		pool:           workerpool.New(poolSize),
-		nextServer:     nextServer,
-		httpServerFQDN: httpServerFQDN,
+		pool:         workerpool.New(poolSize),
+		nextServer:   nextServer,
+		ipxeBaseURL:  ipxeBaseURL,
+		bootsBaseURL: bootsBaseURL,
```
jacobweinstock, line 32: would naming this something that speaks to the ipxe script nature make sense? like `ipxeScriptURL` or `ipxeScriptBase` or similar?

ScottGarman: No objection here. I have a slight preference for ipxeBinariesURL, but not a huge one.

jacobweinstock: nit: use singulars (as these values should only every be a singular location/name).
`bins` -> `bin`, `binaries`-> `binary`
`scripts` -> `script`

jacobweinstock (approved):

cmd/boots/dhcp.go
nshalman: They are singular `URL`s that hold multiple binaries and/or scripts...
Thus my original `xxxxBaseURL`

nshalman: Maybe the existing names are good enough...

PR #254, opened 2022-04-17
Allow registry, registry username and password to be empty strings

jacobweinstock:
## Description

Allow registry, registry username, and password to be empty strings. These values are passed as kernel command line args for OSIE's boot up and then used by Tink worker when it pulls and runs containers. With [this](https://github.com/tinkerbell/tink/pull/607) change to Tink worker, allowing the empty string here provides Tink worker the ability to specify any fully qualified container image in a template. e.g. `docker.io/alpine`, `quay.io/tinkerbell-actions/cexec`, `public.ecr.aws/l0g8r8j6/tinkerbell/pbnj`, etc

This change is backward compatible and shouldn't negatively affect using the Cacher backend.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/installers/osie/main.go
+++ b/installers/osie/main.go
@@ -114,12 +114,18 @@ func kernelParams(ctx context.Context, action, state string, j job.Job, s ipxe.S
 
 	if j.CanWorkflow() {
 		buildWorkerParams()
-		s.Args("docker_registry=" + dockerRegistry)
+		if dockerRegistry != "" {
+			s.Args("docker_registry=" + dockerRegistry)
+		}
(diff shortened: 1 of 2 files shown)


mmlb (commented):
The changes lgtm. I'm wondering we don't go ahead and make it so the kernel args aren't populated if the strings are empty?

jacobweinstock:
> The changes lgtm. I'm wondering we don't go ahead and make it so the kernel args aren't populated if the strings are empty?

sure, i can add that.

mmlb (approved):

PR #255, opened 2022-04-18
Refactor data model

micahhausler:
Signed-off-by: Micah Hausler <mhausler@amazon.com>

## Description

I refactored the boots data model switch. There are now more tree-like call flows without relying on package globals.

## Why is this needed

When adding the Kubernetes data model to Boots, it became clear that much of boots relied on package globals and using JSON as a data format for passing arguments. With this refactor, there are now separate structs with separate responsibilities.

For backend operations there are several new types to reason about.

### Finders
From `client/client.go`
```go
type Finder interface {
	ByIP(context.Context, net.IP) (Discoverer, error)
	ByMAC(context.Context, net.HardwareAddr, net.IP, string) (Discoverer, error)
}
```
which allows for separate implementations of hardware discovery from Packet, Tinkerbell, and Standalone. (You can find the respective finders in `client/{cacher,standalone,tinkerbell}/finder.go`.

### Reporter
Defined in `client/reporter.go`, a Reporter is able to reach out to a server and report state. This was mostly formerly the `Client` type in `packet/client.go`. The former implementations are largely unchanged.
```go
type Reporter interface {
	PostHardwareComponent(ctx context.Context, hardwareID HardwareID, body io.Reader) (*ComponentsResponse, error)
	PostHardwareEvent(ctx context.Context, id string, body io.Reader) (string, error)
	PostHardwarePhoneHome(ctx context.Context, id string) error
	PostHardwareFail(ctx context.Context, id string, body io.Reader) error
	PostHardwareProblem(ctx context.Context, id HardwareID, body io.Reader) (string, error)

	PostInstancePhoneHome(context.Context, string) error
	PostInstanceEvent(ctx context.Context, id string, body io.Reader) (string, error)
	PostInstanceFail(ctx context.Context, id string, body io.Reader) error
	PostInstancePassword(ctx context.Context, id, pass string) error
	UpdateInstance(ctx context.Context, id string, body io.Reader) error
}
```

The interface is currently in the `client/` package, and the implementation in `client/packet`, but I'm open to suggestions for an alternate location, such as `reporter` and `reporter/packet`.

### Discover{er,y}

The former `Discovery` interface has been renamed `Discoverer` to conform to [the Go convention](https://go.dev/doc/effective_go#interface-names) of calling an interface a "verb-er". The `func NewDiscovery(b []byte) (Discovery, error) ` function that switched on an environment variable for the data model is no more. The `Finder` types return the correct `Discoverer` implementations based on their source.

Given that this has less to do with "discovery", I'm up for calling it a different, more specific name.

### Job Creator/Manager

The `job/` package relied on several globals, and no longer does. The new `Creator` type in `job/job.go` is instantiated with a `Finder` and `Reporter` which perform necessary lookups. A `Manager` interface is provided for callers to abstract and test without the implementation.

### BootsDHCPServer and BootsHTTPServer

Similarly to the JobManager, new types in the `cmd/boots` package have been added for DHCP and HTTP serving that contain a `Finder` and `Reporter`.

## How Has This Been Tested?

Only local unit tests. E2E tests forthcoming

Existing users should not observe or need to make any changes.

## Checklist:

I have:

- [x] added unit or e2e tests


client/cacher/finder.go
```diff
@@ -0,0 +1,63 @@
...
+	"encoding/json"
+	"net"
+
+	cacherClient "github.com/packethost/cacher/client"
+	"github.com/packethost/cacher/protos/cacher"
+	"github.com/pkg/errors"
+	"github.com/tinkerbell/boots/client"
+)
+
+type CacherFinder struct {
```
jacobweinstock, line 14: Might not need to include `Cacher` in names in this file with the package being called `cacher`.
prefer `cacher.Finder` over `cacher.CacherFinder`

_"All references to names in your package will be done using the package name, so you can omit that name from the identifiers."_ --[ref](https://github.com/golang/go/wiki/CodeReviewComments#package-names)

client/packet/reporter.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/pkg/errors"
+	"github.com/tinkerbell/boots/client"
+	"github.com/tinkerbell/boots/httplog"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+var _ client.Reporter = &PacketReporter{}
+
+// client has all the fields corresponding to connection
+type PacketReporter struct {
```
jacobweinstock, line 23: Might not need to include `Packet` in names in this file with the package being called `packet`.
prefer `packet.Reporter` over `packet.PacketReporter`

"All references to names in your package will be done using the package name, so you can omit that name from the identifiers." --[ref](https://github.com/golang/go/wiki/CodeReviewComments#package-names)

job/job.go
```diff
@@ -3,52 +3,71 @@ package job
...
 	"github.com/pkg/errors"
+	"github.com/tinkerbell/boots/client"
 	"github.com/tinkerbell/boots/conf"
 	"github.com/tinkerbell/boots/dhcp"
-	"github.com/tinkerbell/boots/packet"
-	tw "github.com/tinkerbell/tink/protos/workflow"
 	"go.opentelemetry.io/otel/trace"
 )
 
+type JobManager struct {
```
jacobweinstock, line 17: Might not need to include `Job` in names in this file with the package being called `job`.
prefer `job.Manager` over `job.JobManager`

"All references to names in your package will be done using the package name, so you can omit that name from the identifiers." --[ref](https://github.com/golang/go/wiki/CodeReviewComments#package-names)

jacobweinstock (changes requested):
thanks for the refactor! I'm stilling going through this, but i put in a few comments after my initial read through.

client/cacher/finder.go
micahhausler: Fixed!

job/job.go
micahhausler: Fixed

client/packet/endpoints.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil, err
+	}
+
+	return &response, nil
+}
+func (c *Reporter) PostHardwareEvent(ctx context.Context, id string, body io.Reader) (string, error) {
+	var res struct {
+		ID string `json:"id"`
+	}
+	if err := c.Post(ctx, "/hardware/"+id+"/events", mimeJSON, body, &res); err != nil {
```
chrisdoherty4, line 75: Relatively minor note, its generally better to use `fmt.Sprintf` or `strings.Join` for concatenation as it results in less allocations stemming from the fact strings are immutable.

client/tinkerbell/finder.go
```diff
@@ -0,0 +1,139 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"net"
+
+	"github.com/pkg/errors"
+	"github.com/tinkerbell/boots/client"
+	tinkClient "github.com/tinkerbell/tink/client"
+	tpkg "github.com/tinkerbell/tink/pkg"
```
chrisdoherty4, line 11: 🤦🏻

client/client.go
```diff
@@ -0,0 +1,91 @@
...
+	Hostname() (string, error)
+	Hardware() Hardware
+	SetMAC(mac net.HardwareAddr)
+}
+
+// HardwareState is the hardware state (e.g. provisioning)
+type HardwareState string
+
+// Hardware interface holds primary hardware methods
+type Hardware interface {
```
chrisdoherty4, line 70: This interface has a bit of a smell about it. I can't quite put my finger on it, but I think its the combination of mac parameterized and non-parmaterized methods. Why does a piece of hardware not know its mac and consequently need a mac for returning things like a kernel path etc?

client/reporter.go
```diff
@@ -0,0 +1,20 @@
+package client
+
+import (
+	"context"
+	"io"
+)
+
+type Reporter interface {
```
chrisdoherty4, line 8: There seems to be a logical separation between these method sets so perhaps they should be distinct interfaces to create granularity? They can always be composed into a super interface providing this complete set.

client/reporter.go
```diff
@@ -0,0 +1,20 @@
...
+	PostHardwareEvent(ctx context.Context, id string, body io.Reader) (string, error)
+	PostHardwarePhoneHome(ctx context.Context, id string) error
+	PostHardwareFail(ctx context.Context, id string, body io.Reader) error
+	PostHardwareProblem(ctx context.Context, id HardwareID, body io.Reader) (string, error)
+
+	PostInstancePhoneHome(context.Context, string) error
+	PostInstanceEvent(ctx context.Context, id string, body io.Reader) (string, error)
+	PostInstanceFail(ctx context.Context, id string, body io.Reader) error
+	PostInstancePassword(ctx context.Context, id, pass string) error
+	UpdateInstance(ctx context.Context, id string, body io.Reader) error
```
chrisdoherty4, line 19: Why does a reporter perform a write operation?

client/client.go
micahhausler: This was moved without changes from `/packet/models.go`. For this PR, I'm not trying to refactor all the business logic, really just focusing on the control flow of how data is fetched from external sources

client/packet/endpoints.go
micahhausler: Agree, again this was ported without changes

client/reporter.go
micahhausler: This is mostly based on the former `Client` interface from `/packet/client.go`. I'm not 100% sure all the places these calls are made, but you might be right. For this PR, I didn't want to change the packet API calls as much as I could, so I might put that into a future PR

client/tinkerbell/finder.go
micahhausler: copied from `packet/endpoints.go`, that wrapper should probably move to some subpackage of pkg

client/reporter.go
micahhausler: No idea.

client/client.go
chrisdoherty4: Gotcha, I didn't make the connection.

client/tinkerbell/finder.go
chrisdoherty4: I just thought it was comical :D

jacobweinstock (dismissed):

mmlb:
So far everything looks ok to me, but I'd like to re-read it again and let it percolate in the brain just a little bit :D.

mmlb:
Ok started on v2 :eyes:. Trying this locally I'm seeing that `client/tinkerbell/mock_hardware/hardware_mock.go` has some modifications that need to be checked in?

```
go test ./...
# github.com/tinkerbell/boots/client/tinkerbell/mock_hardware
client/tinkerbell/mock_hardware/hardware_mock.go:160:118: undefined: hardware.HardwareService_WatchClient
client/tinkerbell/mock_hardware/hardware_mock.go:167:21: undefined: hardware.HardwareService_WatchClient
```

jacobweinstock:
> Ok started on v2 👀. Trying this locally I'm seeing that `client/tinkerbell/mock_hardware/hardware_mock.go` has some modifications that need to be checked in?
> 
> ```
> go test ./...
> # github.com/tinkerbell/boots/client/tinkerbell/mock_hardware
> client/tinkerbell/mock_hardware/hardware_mock.go:160:118: undefined: hardware.HardwareService_WatchClient
> client/tinkerbell/mock_hardware/hardware_mock.go:167:21: undefined: hardware.HardwareService_WatchClient
> ```

Hey @mmlb, I believe i have fixed this. Thanks.

jacobweinstock (approved):

mmlb (commented):
I thought I published these on Friday. Still going through the PR.

client/cacher/discovery.go
```diff
@@ -291,6 +289,10 @@ func (h HardwareCacher) Interfaces() []Port {
 	return ports
 }
 
+type InterfaceCacher struct {
```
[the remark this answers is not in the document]
micahhausler: Removed

client/packet/endpoints.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+const mimeJSON = "application/json"
+
+// ReportDiscovery is called when Cacher returns an empty response for the MAC
+// address. It does a POST to the Packet API /staff/cacher/hardware-discovery
+// endpoint.
+// This was split out from DiscoverHardwareFromDHCP to make the control flow
+// easier to understand.
+func (c *Reporter) ReportDiscovery(ctx context.Context, mac net.HardwareAddr, giaddr net.IP, circuitID string) (client.Discoverer, error) {
```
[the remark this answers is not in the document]
micahhausler: Sure will do, although it is doing a POST request

client/cacher/finder_test.go
```diff
@@ -0,0 +1,200 @@
...
+			name: "query error",
+			arg: func() net.HardwareAddr {
+				mac, _ := net.ParseMAC("ab:cd:ef:01:12:34")
+
+				return mac
+			}(),
+			resp:    nil,
+			respErr: errors.New("no hardware"),
+			want:    nil,
+			wantErr: errors.New("get hardware by ip from cacher: no hardware"),
```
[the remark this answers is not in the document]
micahhausler: good catch

client/cacher/finder.go
```diff
@@ -0,0 +1,67 @@
...
+	return d, nil
+}
+
+// ByMAC returns a Discoverer for a particular MAC address.
+func (f *Finder) ByMAC(ctx context.Context, mac net.HardwareAddr, _ net.IP, _ string) (client.Discoverer, error) {
+	resp, err := f.cc.ByMAC(ctx, &cacher.GetRequest{
+		MAC: mac.String(),
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "get hardware by ip from cacher")
```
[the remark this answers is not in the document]
micahhausler: good catch

client/packet/endpoints.go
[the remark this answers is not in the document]
mmlb: Yeah I know, its unfortunate but I'd rather the function name match what its doing semantically instead of a reflection of the transport.

client/packet/endpoints.go
```diff
@@ -0,0 +1,131 @@
...
+	"bytes"
+	"context"
+	"encoding/json"
+	"io"
+	"net"
+
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/tinkerbell/boots/client"
+	bootscacher "github.com/tinkerbell/boots/client/cacher"
```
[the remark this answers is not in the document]
micahhausler: I think this file used to import an external cacher package too, and this was to differentiate. It no longer makes sense to keep the alias

client/packet/noop_reporter.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (c *noOpReporter) PostInstancePassword(ctx context.Context, id, pass string) error {
+	return nil
+}
+
+func (c *noOpReporter) UpdateInstance(ctx context.Context, id string, body io.Reader) error {
+	return nil
+}
+
+// NewNoOpReporter returns a reporter that does nothing.
```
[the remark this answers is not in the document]
micahhausler: yea, thanks

client/standalone/discover.go
```diff
@@ -0,0 +1,77 @@
+package standalone
+
+import (
+	"net"
+	"time"
+
+	"github.com/tinkerbell/boots/client"
+)
+
+// discover.go contains a standalone backend for boots so it can run without
```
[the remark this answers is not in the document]
micahhausler: yes

client/packet/noop_reporter.go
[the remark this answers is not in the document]
chrisdoherty4: Isn't that just "Where its useful". We could mentioned tests but do code docs really need to list where to use something as open as this?

client/tinkerbell/discovery.go
```diff
@@ -21,9 +23,13 @@ type DiscoveryTinkerbellV1 struct {
...
 type HardwareTinkerbellV1 struct {
-	ID       string   `json:"id"`
-	Network  Network  `json:"network"`
-	Metadata Metadata `json:"metadata"`
+	ID       string          `json:"id"`
+	Network  client.Network  `json:"network"`
+	Metadata client.Metadata `json:"metadata"`
+}
+
+type InterfaceTinkerbell struct {
```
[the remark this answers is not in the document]
micahhausler: yep good catch

job/job.go
```diff
@@ -3,52 +3,65 @@ package job
...
-func SetClient(c packet.Client) {
-	client = c
+// NewCreator returns a manager that can create jobs
+func NewCreator(logger log.Logger, provisionerEngineName string, reporter client.Reporter, finder client.Finder) *Creator {
+	return &Creator{
+		finder:                finder,
+		provisionerEngineName: provisionerEngineName,
+		reporter:              reporter,
+		logger:                logger,
+		nowFunc:               time.Now,
```
mmlb, line 39: Why do we want a nowFunc here? Its only ever set to `time.Now` so why the indirection?

client/packet/noop_reporter.go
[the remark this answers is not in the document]
micahhausler: I forgot to add this to `cmd/main.go` for non-cacher backends. Rather than refactor every instance where reporting is touched, I thought it might be easier to have a no-op and later refactor that

client/client.go
```diff
@@ -0,0 +1,91 @@
...
+
+var (
+	ErrNotFound = errors.New("hardware not found")
+)
+
+// Finder is a type for discovering hardware.
+type Finder interface {
+	ByIP(context.Context, net.IP) (Discoverer, error)
+	ByMAC(context.Context, net.HardwareAddr, net.IP, string) (Discoverer, error)
+}
```
mmlb, line 19: Can you do `HardwareFinder` instead of bare `Finder` please, if we've got 2 of them then I think it makes most sense to have them easier to identify by type/name.

micahhausler: Sure

client/packet/noop_reporter.go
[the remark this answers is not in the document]
mmlb: @chrisdoherty4 normally yeah, but documenting when/why someone would want one seems most useful for noop vs EM, or tink or w/e else because its kind of obvious from the name (but being explicit in the documentation doesn't hurt for those cases either).

job/job.go
micahhausler: For testability, but a quick search shows that nothing references this, I'll change it back

mmlb (approved):
thanks for the work, this is a great step forward!

PR #256, opened 2022-04-19
Add kernel commandline parameter for specifying the tink-worker image:

jacobweinstock:
## Description

This allows flexibility from where to pull the tink worker container image. This will be needed if the env var `DOCKER_REGISTRY` is not specified.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/boots/http.go
+++ b/cmd/boots/http.go
@@ -55,15 +55,16 @@ func otelFuncWrapper(route string, h func(w http.ResponseWriter, req *http.Reque
 }
 
 type jobHandler struct {
-	i job.Installers
+	i               job.Installers
+	tinkWorkerImage string
 }
 
 // ServeHTTP sets up all the HTTP routes using a stdlib mux and starts the http
 // server, which will block. App functionality is instrumented in Prometheus and
 // OpenTelemetry. Optionally configures X-Forwarded-For support.
-func ServeHTTP(i job.Installers, addr string, ipxePattern string, ipxeHandler func(http.ResponseWriter, *http.Request)) {
+func ServeHTTP(i job.Installers, addr, tinkWorkerImage string, ipxePattern string, ipxeHandler func(http.ResponseWriter, *http.Request)) {
 	mux := http.NewServeMux()
-	s := jobHandler{i: i}
+	s := jobHandler{i: i, tinkWorkerImage: tinkWorkerImage}
 	mux.Handle(otelFuncWrapper("/", s.serveJobFile))
 	if ipxeHandler != nil {
 		mux.Handle(otelFuncWrapper(ipxePattern, ipxeHandler))
@@ -148,6 +149,8 @@ func (h *jobHandler) serveJobFile(w http.ResponseWriter, req *http.Request) {
 
 		return
 	}
+	// set Tink Worker image location
+	j.TinkWorkerImage = h.tinkWorkerImage
 	// This gates serving PXE file by
 	// 1. the existence of a hardware record in tink server
 	// AND
--- a/cmd/boots/main.go
+++ b/cmd/boots/main.go
@@ -76,6 +76,8 @@ type config struct {
 	syslogAddr string
 	// loglevel is the log level for boots
 	logLevel string
+	// tinkWorkerImage is used in kernel command line parameters to define where the Tink worker image is located.
+	tinkWorkerImage string
 }
 
 func main() {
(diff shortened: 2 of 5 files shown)


mmlb:
This pr is alright[^1] but I kinda want to push back on it in that I feel like we shouldn't be specifying the worker to use. Letting that change due to a boots update/deployment creates a coupling that doesn't really make sense.

[^1]: my only nit for the code at hand is that `location` shouldn't be part of the comments, instead just using `tink-worker image`

jacobweinstock:
> This pr is alright[1](#user-content-fn-1-ef92e2e84079dc877c4e8dc3fe84f30b) but I kinda want to push back on it in that I feel like we shouldn't be specifying the worker to use. Letting that change due to a boots update/deployment creates a coupling that doesn't really make sense.
> 
> ## Footnotes
> 1. my only nit for the code at hand is that `location` shouldn't be part of the comments, instead just using `tink-worker image` [↩](#user-content-fnref-1-ef92e2e84079dc877c4e8dc3fe84f30b)

Sorry, I dont know if i follow. the env `DOCKER_REGISTRY` already allows us the ability to specify from where to pull the tink worker. How would this be any different from that? This just decouples `DOCKER_REGISTRY` from the tink worker location so that an empty `DOCKER_REGISTRY` doesnt break. See bootkit in hook [here](https://github.com/tinkerbell/hook/blob/16a89f5fbdf1be008e4199407fcb24bd388cc1be/bootkit/main.go#L64)

mmlb:
> Sorry, I dont know if i follow. the env `DOCKER_REGISTRY` already allows us the ability to specify from where to pull the tink worker. How would this be any different from that? This just decouples `DOCKER_REGISTRY` from the tink worker location so that an empty `DOCKER_REGISTRY` doesnt break. See bootkit in hook [here](https://github.com/tinkerbell/hook/blob/16a89f5fbdf1be008e4199407fcb24bd388cc1be/bootkit/main.go#L64)

Right but boots isn't prescribing which tink-worker image should be used, this PR changes that. Boots should not gain this new policy/ability. The image being booted should either know what tink-worker it wants to use (because its been tested with it), or it should know that it wants to use :latest or maybe it figures that out from some other means.

jacobweinstock:
> > Sorry, I dont know if i follow. the env `DOCKER_REGISTRY` already allows us the ability to specify from where to pull the tink worker. How would this be any different from that? This just decouples `DOCKER_REGISTRY` from the tink worker location so that an empty `DOCKER_REGISTRY` doesnt break. See bootkit in hook [here](https://github.com/tinkerbell/hook/blob/16a89f5fbdf1be008e4199407fcb24bd388cc1be/bootkit/main.go#L64)
> 
> Right but boots isn't prescribing which tink-worker image should be used, this PR changes that. Boots should not gain this new policy/ability. The image being booted should either know what tink-worker it wants to use (because its been tested with it), or it should know that it wants to use :latest or maybe it figures that out from some other means.

"Boots should not gain this new policy/ability." Sorry, I'm not seeing why this should be the case. There is already "coupling" as you say between many Boots env vars -> cmdline args -> bootkit.

Why does hard-coding the image name `tink-worker` matter?

Also, this change is optional and does not change the existing behavior. So the previous behavior of `tink-worker:latest` in conjunction with a specific registry is still available.

mmlb:
> > > Sorry, I dont know if i follow. the env `DOCKER_REGISTRY` already allows us the ability to specify from where to pull the tink worker. How would this be any different from that? This just decouples `DOCKER_REGISTRY` from the tink worker location so that an empty `DOCKER_REGISTRY` doesnt break. See bootkit in hook [here](https://github.com/tinkerbell/hook/blob/16a89f5fbdf1be008e4199407fcb24bd388cc1be/bootkit/main.go#L64)
> > 
> > 
> > Right but boots isn't prescribing which tink-worker image should be used, this PR changes that. Boots should not gain this new policy/ability. The image being booted should either know what tink-worker it wants to use (because its been tested with it), or it should know that it wants to use :latest or maybe it figures that out from some other means.
> 
> "Boots should not gain this new policy/ability." Sorry, I'm not seeing why this should be the case. There is already "coupling" as you say between many Boots env vars -> cmdline args -> bootkit.

Because boots doesn't need to provide the tink worker image arg in order to get its job done, to boot an image. We've been overloading stuff into boots because its convenient but we shouldn't continue to add on top. Solving this outside of boots could also open up the space to solve other use cases (I'm thinking of what we discussed in the last contributor's meeting about ssh-keys and an external url to fetch things).

IMO we should be working towards removing these. I can imagine doing this right and having just 1 cmdline argument for "fetch things you need to know from this url", where we can provide registry info & creds, images to use, extra services to enable maybe...

> 
> Why does hard-coding the image name `tink-worker` matter?
> 
> Also, this change is optional and does not change the existing behavior. So the previous behavior of `tink-worker:latest` in conjunction with a specific registry is still available.

It just doesn't seem like a good enough benefit compared to the cost. It doesn't help out with wanting to roll out new versions of tink-worker safely if it means we need a redeployment of boots and doing so now makes it live for everything boot that will go through boots. Solving the "how to get inputs into the live environment" outside of boots is better done outside of boots.

Optional features are worse than required ones imo, because now you have to account for both scenarios (I'm keenly aware of this after going through and removing all the mirror stuff from conf/mirror.go in #258).

jacobweinstock:
Hey @mmlb, thanks for the back and forth. I'm going to close this one and see about solving this a different way.

PR #257, opened 2022-04-19
pass tink server tls mode as kernel cmdline arg

jacobweinstock:
## Description

This allows TLS communication with Tink server to be toggled by Tink worker. 

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


mmlb:
This is now in main via #258 see https://github.com/tinkerbell/boots/blob/main/installers/osie/main.go#L65. Closing.

PR #250, opened 2022-04-06
Add kubernetes-based backend

micahhausler:
Signed-off-by: Micah Hausler <mhausler@amazon.com>

## Description

The new backend uses the Kubernetes CRD types in Tinkerbell core. Boots could use a bit of a refactor/organization in how the backends work, but for now I've tried to fit in to the existing structure. 

## Why is this needed

This is the boots side of work for tinkerbell/proposals#46

Fixes: #

## How Has This Been Tested?

Tested with real bare metal workers using an image built from this branch. I will add E2E tests in a follow up PR, or can add to this branch

## How are existing users impacted? What migration steps/scripts do we need?

No impact to existing users

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required)
- [ ] added unit or e2e tests
- [ ] provided instructions on how to upgrade

PR #258, opened 2022-04-22
Spring cleaning

mmlb:
## Description

I was getting ready to PR my branch that is updated to work in /cert-less tink (now that https://github.com/tinkerbell/tink/pull/584 is merged) and it was basically a situation where pulling one thread unraveled a bunch of other things. Dropping /cert begot some refactors (started by dropping the `buildWorkerParams` silliness) which begot some test fixes which begot some other refactors which was also dumb to do for rancher, nixos, and centos. So here we are.

## Why is this needed

Makes updating to post https://github.com/tinkerbell/tink/pull/584 easier/nicer.

## How Has This Been Tested?

Ran unit tests.

Less EM specific junk is always good. Less duplicate/unclear tests too. This PR does introduce a change wrt sandbox, where MIRROR_HOST is dropped in favor of MIRROR_BASE_URL, I'll follow up with a PR there once this merges.

## Checklist:

I have:

- [ ] added unit or e2e tests


micahhausler (approved):
LGTM, but I only reviewed for language formatting and idioms. I'd want another set of eyes for PXE content

micahhausler:
@mmlb do you think we can get #255 in first?

mmlb:
Hey @jacobweinstock yeah I recognize this is a big PR. I kind of went that way to avoid a long series of small quick PRs when effectively they are all refactors/tidying up of things. I can definitely split up into smaller PRs but honestly I think it'd be easier to go through this commit by commit, each should be pretty self-explanatory/stand-alone and digestable. Let me know if that works or not. #255 is close so to ready and I'll be rebasing on top of it.

go.mod
```diff
@@ -22,46 +22,169 @@ require (
...
+	golang.org/x/crypto v0.0.0-20220214200702-86341886e292
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	golang.org/x/tools v0.1.10
-	google.golang.org/genproto v0.0.0-20211021150943-2b146023228c // indirect
+	google.golang.org/genproto v0.0.0-20211208223120-3a66f561d7aa // indirect
 	google.golang.org/grpc v1.44.0
 	inet.af/netaddr v0.0.0-20211027220019-c74959edd3b6
 )
 
+require github.com/golangci/golangci-lint v1.45.2
```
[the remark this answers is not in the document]
mmlb: The Makefile has a target that runs golangci-lint and says `@echo be sure golangci-lint is installed: https://golangci-lint.run/usage/install/` which really makes no sense when we are pinning go tools using `tools.go`.

conf/mirror.go
```diff
@@ -2,97 +2,39 @@ package conf
...
-
-		return u, nil
-	}
-	u, err := base.Parse(defaultMirrorPath)
-	if err != nil {
-		return nil, errors.Wrapf(err, "invalid default mirror path %q", defaultMirrorPath)
+func mustBuildMirrorBaseURL() string {
+	s, err := buildMirrorBaseURL()
+	if s == "" {
+		panic(fmt.Sprintf(`conf: MirrorBaseURL: %q: %s\n`, s, err))
```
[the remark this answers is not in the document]
mmlb: this was already here from before, I've actually removed one other case of unwanted behavior. I'll rework the code though so that we don't need the conf package anymore.

installers/osie/main.go
```diff
@@ -4,87 +4,122 @@ import (
...
-	return func(ctx context.Context, j job.Job, s ipxe.Script) ipxe.Script {
-		if j.Rescue() {
-			return i.rescue()(ctx, j, s)
+	staticTinkArgs []string
+)
+
+type installer struct{}
+
+func Installer() job.BootScripter {
+	if env.Get("DATA_MODEL_VERSION") != "1" {
```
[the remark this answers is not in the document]
mmlb: I've lifted this up to here from where it used to be, this env var was being checked for *every* ipxe script that was generated (https://github.com/tinkerbell/boots/blob/main/installers/osie/main.go#L115-L116) instead of just once at boot since they aren't going to change. I'll move this to be done in cmd/boots/main.go and pass it in as args.

installers/osie/main.go
```diff
@@ -4,87 +4,122 @@ import (
...
-func (i Installer) rescue() job.BootScript {
-	return func(ctx context.Context, j job.Job, s ipxe.Script) ipxe.Script {
-		s.Set("action", "rescue")
-		s.Set("state", j.HardwareState())
+	staticTinkArgs = []string{
+		"grpc_authority=" + require("TINKERBELL_CERT_URL"),
+		"grpc_cert_url=" + require("TINKERBELL_GRPC_AUTHORITY"),
+		"packet_base_url=" + workflowURL,
+	}
+	if registry := env.Get("DOCKER_REGISTRY"); registry != "" {
```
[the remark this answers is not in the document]
mmlb: similar to the other comment and I'll rework to avoid this now

installers/osie/main.go
```diff
@@ -4,87 +4,122 @@ import (
...
-}
 
-// rescue generates the ipxe boot script for booting into osie in rescue mode
-func (i Installer) rescue() job.BootScript {
-	return func(ctx context.Context, j job.Job, s ipxe.Script) ipxe.Script {
-		s.Set("action", "rescue")
-		s.Set("state", j.HardwareState())
+	staticTinkArgs = []string{
+		"grpc_authority=" + require("TINKERBELL_CERT_URL"),
+		"grpc_cert_url=" + require("TINKERBELL_GRPC_AUTHORITY"),
```
micahhausler, line 41: Mix up here:
```suggestion
		"grpc_authority=" + require("TINKERBELL_GRPC_AUTHORITY"),
		"grpc_cert_url=" + require("TINKERBELL_CERT_URL"),
```

Also:
* Isn't `TINKERBELL_CERT_URL` deprecated/gone and replaced with `TINKERBELL_TLS=bool`?
* You'll need a way to plumb down the tink API for the Kubernetes backend data model

rules.mk
```diff
@@ -39,23 +39,27 @@ endif
...
 	syslog/severity_string.go \
-	
-.PHONY: $(generated_go_files)
 
 # go generate
 go_generate:
 $(filter %_string.go,$(generated_go_files)): bin/stringer
 $(filter %_mock.go,$(generated_go_files)): bin/mockgen
+packet/mock_cacher/cacher_mock.go: packet/models_cacher.go
+packet/mock_workflow/workflow_mock.go: packet/models_tinkerbell.go
```
micahhausler, line 60: These paths need updating
```suggestion
client/cacher/mock_cacher/cacher_mock.go: client/cacher/discovery.go
client/tinkerbell/mock_workflow/workflow_mock.go: client/tinkerbell/discovery.go
client/tinkerbell/mock_hardware/hardware_mock.go: client/tinkerbell/discovery.go
```

micahhausler (changes requested):
Reviewed for go style and backend support, I will defer to other reviewers for most of the content

conf/mirror.go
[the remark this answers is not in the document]
mmlb: I tried re-working but then came across all the other things in https://github.com/tinkerbell/boots/blob/main/conf/config.go and stopped trying. I've made it so osie is better about this as a small step in the right direction. Getting rid of all the `must*` in conf is going to explode this PR even further.

installers/osie/main.go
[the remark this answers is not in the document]
mmlb: fixed.

installers/osie/main.go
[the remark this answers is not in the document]
mmlb: fixed

installers/osie/main.go
mmlb: oops I've *still* forgotten about k8s backend, have to update that too.

mmlb: I kept CERT_URL and left TLS out originally because it didn't seem like "spring cleaning", but after you mentioned it I think its fine, it should have been done when those respective PRs landed in tink so I can see this being "dirty" in need of cleaning ... if you squint :D. So I've fixed this now.

rules.mk
mmlb: fixed, good catch.

installers/osie/main.go
mmlb: @micahhausler the kubernetes backend pr (#250) isn't merged yet.

go.mod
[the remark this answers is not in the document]
jacobweinstock: the other imports in tools.go don't show up in the main.go. i wondering why golangci-lint does?

installers/osie/main.go
```diff
@@ -10,99 +10,138 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-type Installer struct {
-	// ExtraKernelArgs are key=value pairs to be added as kernel commandline to the kernel in iPXE for OSIE.
-	ExtraKernelArgs string
+type installer struct {
+	osieURL string
+
```
jacobweinstock, line 15: nit: remove extra space

installers/osie/main.go
```diff
@@ -111,30 +150,17 @@ func (i Installer) kernelParams(ctx context.Context, action, state string, j job
...
 	}
 
 	if j.CanWorkflow() {
-		buildWorkerParams()
-		if dockerRegistry != "" {
-			s.Args("docker_registry=" + dockerRegistry)
-		}
-		s.Args("grpc_authority=" + grpcAuthority)
-		s.Args("grpc_cert_url=" + grpcCertURL)
+		s.Args(i.tinkParams)
```
jacobweinstock, line 162: in my tests this appears to add a single space if `i.tinkParams == ""`.

cmd/boots/main.go
```diff
@@ -363,38 +361,46 @@ func newCLI(cfg *config, fs *flag.FlagSet) *ffcli.Command {
...
-	// register osie as default
-	d := osie.Installer{ExtraKernelArgs: cf.extraKernelArgs}
-	i.RegisterDefaultInstaller(d.DefaultHandler())
-	// register rancher
-	r := rancher.Installer{}
-	i.RegisterDistro("rancher", r.BootScript())
+	dataModelVersion := env.Get("DATA_MODEL_VERSION")
+	auth := env.Get("TINKERBELL_GRPC_AUTHORITY")
+	if dataModelVersion == "1" && auth == "" {
+		mainlog.Fatal(errors.New("TINKERBELL_GRPC_AUTHORITY is required when in tinkerbell mode"))
```
jacobweinstock, line 378: Can this panic be moved to `func main`?

go.mod
```diff
@@ -11,7 +11,8 @@ require (
 	github.com/go-logr/logr v1.2.2
 	github.com/go-logr/zapr v1.2.2
 	github.com/golang/mock v1.6.0
-	github.com/google/go-cmp v0.5.6
+	github.com/golangci/golangci-lint v1.45.2
```
jacobweinstock, line 14: hmm....still confused why this is needed here.

mmlb: nope they're here too:
```
[10:14:49]-[~/go/src/github.com/tinkerbell/boots]─[manny@zennix]> 
cat tools.go 
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: tools.go
       │ Size: 228 B
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ //go:build tools
   2   │ // +build tools
   3   │ 
   4   │ package tools
   5   │ 
   6   │ import (
   7   │     _ "github.com/golang/mock/mockgen"
   8   │     _ "github.com/golangci/golangci-lint/cmd/golangci-lint"
   9   │     _ "golang.org/x/tools/cmd/goimports"
  10   │     _ "golang.org/x/tools/cmd/stringer"
  11   │ )
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
                                                                                                                                                                      
[10:14:51]-[~/go/src/github.com/tinkerbell/boots]─[manny@zennix]> 
grep -e golang/mock -e golang.org/x/tools -e golangci-lint go.mod 
go.mod:    github.com/golang/mock v1.6.0
go.mod:    github.com/golangci/golangci-lint v1.45.2
go.mod:    golang.org/x/tools v0.1.10
```

installers/osie/main.go
mmlb: tinkParams will never be empty since its initialized with `grpc_authority` and `packet_base_url`.

jacobweinstock: if `dataModelVersion != "1"` then `i.tinkParams == ""`

mmlb: hmm good point, this is still guarded by `CanWorkflow` which doesn't make any sense when dataModelVersion != 1. :thinking: ideally CanWorkflow returns false if dataModelVersion != 1 always, let me try that out.

installers/osie/main.go
```diff
@@ -10,99 +10,138 @@ import (
...
+	}
+	if registry != "" {
+		tinkParams = append(tinkParams, "docker_registry="+registry)
+	}
+	if registryUsername != "" {
+		tinkParams = append(tinkParams, "registry_username="+registryUsername)
+	}
+	if registryPassword != "" {
+		tinkParams = append(tinkParams, "registry_password="+registryPassword)
+	}
```
jacobweinstock, line 72: I believe these would still be needed even if `dataModelVersion != "1"`

mmlb: thats not how the code is today: https://github.com/tinkerbell/boots/blob/main/installers/osie/main.go#L123-L139. This is behind `CanWorkflow` but that only ever happens when dataModelVersion == 1 anyway. We shouldn't have 2 ways to do this check.

installers/osie/main.go
mmlb: I'm going to make sure we're in tink mode as a prereq for the CanWorkflow checks. Will also open up an issue so that we drop CanWorkflow all together as boots doesn't really support a mixed-mode operation we're either in cacher, tink, standalone and the hw has no real say in it.

installers/osie/main.go
jacobweinstock: thats not my experience. i can use the `DATA_MODEL_VERSION=standalone` and still use `CanWorkflow`.

mmlb: hmm this seems very accidental and confusing, if you're in dmv=standalone and canworkflow=true then why would tink connection params matter? dmv is actually doing 2 different jobs and causing confusion (surprise, surprise... :smile:). `DATA_SOURCE=cacher|tink|standalone|something-in-the-future/pluggable` DATA_MODEL=cacher|tink`. But oh well, not going to try and do that now. I'll fix it so tinkArgs isn't dependent on DMV=1.

installers/osie/main.go
mmlb: I believe this has been fixed now, can you make sure please?

cmd/boots/main.go
mmlb: done

installers/osie/main.go
mmlb: done

jacobweinstock (approved):
Thanks @mmlb !

PR #261, opened 2022-04-29
Allow extra kernel args, set via cli flag:

jacobweinstock:
## Description

This allows operators to set any necessary and needed kernel commandline args to the default installer; OSIE. This makes custom kernel args a first class option instead of the current model to overload the "facility" field in the backend data.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/boots/main.go
+++ b/cmd/boots/main.go
@@ -347,7 +347,7 @@ func newCLI(cfg *config, fs *flag.FlagSet) *ffcli.Command {
 	fs.StringVar(&cfg.logLevel, "log-level", "info", "log level.")
 	fs.StringVar(&cfg.dhcpAddr, "dhcp-addr", conf.BOOTPBind, "IP and port to listen on for DHCP.")
 	fs.StringVar(&cfg.syslogAddr, "syslog-addr", conf.SyslogBind, "IP and port to listen on for syslog messages.")
-	fs.StringVar(&cfg.extraKernelArgs, "extra-kernel-args", "", "Extra set of kernel args (k=v k=v) that are appended to the kernel in iPXE for the default iPXE installer.")
+	fs.StringVar(&cfg.extraKernelArgs, "extra-kernel-args", "", "Extra set of kernel args (k=v k=v) that are appended to the kernel cmdline when booting via iPXE.")
 
 	return &ffcli.Command{
 		Name:       name,
--- a/cmd/boots/main.go
+++ b/cmd/boots/main.go
@@ -347,7 +347,7 @@ func newCLI(cfg *config, fs *flag.FlagSet) *ffcli.Command {
 	fs.StringVar(&cfg.logLevel, "log-level", "info", "log level.")
 	fs.StringVar(&cfg.dhcpAddr, "dhcp-addr", conf.BOOTPBind, "IP and port to listen on for DHCP.")
 	fs.StringVar(&cfg.syslogAddr, "syslog-addr", conf.SyslogBind, "IP and port to listen on for syslog messages.")
-	fs.StringVar(&cfg.extraKernelArgs, "extra-kernel-args", "", "Extra set of kernel args (k=v k=v) that are appended to the kernel in iPXE for OSIE.")
+	fs.StringVar(&cfg.extraKernelArgs, "extra-kernel-args", "", "Extra set of kernel args (k=v k=v) that are appended to the kernel in iPXE for the default iPXE installer.")
 
 	return &ffcli.Command{
 		Name:       name,
--- a/cmd/boots/main.go
+++ b/cmd/boots/main.go
@@ -79,6 +79,8 @@ type config struct {
 	syslogAddr string
 	// loglevel is the log level for boots
 	logLevel string
+	// extraKernelArgs are key=value pairs to be added as kernel commandline to the kernel in iPXE for OSIE
+	extraKernelArgs string
 }
 
 func main() {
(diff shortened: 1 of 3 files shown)


cmd/boots/main.go
```diff
@@ -286,6 +288,7 @@ func newCLI(cfg *config, fs *flag.FlagSet) *ffcli.Command {
 	fs.StringVar(&cfg.logLevel, "log-level", "info", "log level.")
 	fs.StringVar(&cfg.dhcpAddr, "dhcp-addr", conf.BOOTPBind, "IP and port to listen on for DHCP.")
 	fs.StringVar(&cfg.syslogAddr, "syslog-addr", conf.SyslogBind, "IP and port to listen on for syslog messages.")
+	fs.StringVar(&cfg.kernelArgs, "kernel-args", "", "Kernel args to add to iPXE initrd for OSIE.")
```
micahhausler, line 291: question: Does a string make sense here or a map of string to string, or a slice of strings?

mmlb, line 291: Can we do something like `extra-kernel-args, ..., "Extra set of kernel args that are appended to iPXE initrd for OSIE"` or something along those lines? I think that will avoid causing confusion for someone just looking around and running `-h`.

jacobweinstock: Unfortunately, i don't believe `*flag.FlagSet` has either of those options.

micahhausler: You can drop in `flag "github.com/spf13/pflag"` as a replacement, and it has both those types

jacobweinstock: updated.

jacobweinstock: I tried to use `github.com/spf13/pflag` but the use of `github.com/peterbourgon/ff/ffcli` requires the stdlib flag package. see [here](https://github.com/peterbourgon/ff/blob/7c05c99406b871976b139e55f9a771f22769f9de/ffcli/command.go#L55).

cmd/boots/main.go
```diff
@@ -375,7 +378,7 @@ func registerInstallers() job.Installers {
 	o := osie.Installer{}
 	i.RegisterDistro("discovery", o.Discover())
 	// register osie as default
-	d := osie.Installer{}
+	d := osie.Installer{ExtraKernelArgs: strings.Split(cf.extraKernelArgs, ",")}
```
mmlb, line 381: it actually doesn't make any sense to split this at all, we don't ever interpret the args and don't care what format it is. I think doing the split will likely just do the wrong thing for args that themselves have a `,` (e.g.: `console=ttyS0,115200n8` or similar)

installers/osie/main.go
```diff
@@ -96,6 +99,13 @@ func kernelParams(ctx context.Context, action, state string, j job.Job, s ipxe.S
 	s.Args("packet_state=${state}")
 	s.Args("osie_vendors_url=" + conf.OsieVendorServicesURL)
 
+	// Add extra kernel args
+	for _, kv := range i.ExtraKernelArgs {
+		if strings.Contains(kv, "=") {
+			s.Args(kv)
+		}
+	}
```
mmlb, line 107: this should probably be a simple `if i.ExtraKernelArgs != "" {` instead.

cmd/boots/main.go
jacobweinstock: ah, great point. will update.

cmd/boots/main.go
```diff
@@ -345,6 +347,7 @@ func newCLI(cfg *config, fs *flag.FlagSet) *ffcli.Command {
 	fs.StringVar(&cfg.logLevel, "log-level", "info", "log level.")
 	fs.StringVar(&cfg.dhcpAddr, "dhcp-addr", conf.BOOTPBind, "IP and port to listen on for DHCP.")
 	fs.StringVar(&cfg.syslogAddr, "syslog-addr", conf.SyslogBind, "IP and port to listen on for syslog messages.")
+	fs.StringVar(&cfg.extraKernelArgs, "extra-kernel-args", "", "Extra set of kernel args (k=v k=v) that are appended to the kernel in iPXE for OSIE.")
```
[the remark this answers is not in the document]
mmlb: I've been thinking of OSIE as small caps, *an* os install env not tinkerbell/osie. We should probably come up with a better name project wide e.g. (https://github.com/tinkerbell/tinkerbell.org/issues/212)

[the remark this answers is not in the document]
jacobweinstock: yeah, I'll update. OSIE was supposed to be a general name (Operating System Installation Environment) for which ever OS was used to run workflows but unfortunately it got coupled with the repo. Regardless, I've updated this help line.

cmd/boots/main.go
jacobweinstock: updated

installers/osie/main.go
jacobweinstock: updated

cmd/boots/main.go
```diff
@@ -345,6 +347,7 @@ func newCLI(cfg *config, fs *flag.FlagSet) *ffcli.Command {
 	fs.StringVar(&cfg.logLevel, "log-level", "info", "log level.")
 	fs.StringVar(&cfg.dhcpAddr, "dhcp-addr", conf.BOOTPBind, "IP and port to listen on for DHCP.")
 	fs.StringVar(&cfg.syslogAddr, "syslog-addr", conf.SyslogBind, "IP and port to listen on for syslog messages.")
+	fs.StringVar(&cfg.extraKernelArgs, "extra-kernel-args", "", "Extra set of kernel args (k=v k=v) that are appended to the kernel in iPXE for the default iPXE installer.")
```
mmlb, line 350: I found the latest wording a little awkward still, how about:

```suggestion
	fs.StringVar(&cfg.extraKernelArgs, "extra-kernel-args", "", "Extra set of kernel args (k=v k=v) that are appended to the kernel cmdline when booting via iPXE.")
```

jacobweinstock: updated

micahhausler (approved):
LGTM

PR #263, opened 2022-05-03
CLI flag to allow overriding of OSIE path

ptrivedi:
Add CLI flag for overriding OSIE path

If an OSIE download path URL is specified via this fkag, OSIE_PATH and
MIRROR_HOST variables that are used to build OSIE URL will be ignored.
Hardcoded appending of 'misc/osie' and 'current' directories to the path
will be bypassed.

With the ongoing refactoring efforts, once all this hardcoded appending 
of paths is cleaned up, this flag can be removed/modified.

Signed-off-by: Pooja Trivedi <tripooja@amazon.com>

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/boots/main.go
+++ b/cmd/boots/main.go
@@ -86,6 +86,9 @@ type config struct {
 	kubeAPI string
 	// kubeNamespace is an override for the namespace the kubernetes client will watch.
 	kubeNamespace string
+	// osiePathOverride allows a completely custom path/URL to be specified for OSIE/Hook images
+	// This will bypass the hardcoded path appending of 'misc/osie/current' to the path
+	osiePathOverride string
 }
 
 func main() {
@@ -375,6 +378,7 @@ func newCLI(cfg *config, fs *flag.FlagSet) *ffcli.Command {
 	fs.StringVar(&cfg.kubeconfig, "kubeconfig", "", "The Kubernetes config file location. Only applies if DATA_MODEL_VERSION=kubernetes.")
 	fs.StringVar(&cfg.kubeAPI, "kubernetes", "", "The Kubernetes API URL, used for in-cluster client construction. Only applies if DATA_MODEL_VERSION=kubernetes.")
 	fs.StringVar(&cfg.kubeNamespace, "kube-namespace", "", "An optional Kubernetes namespace override to query hardware data from.")
+	fs.StringVar(&cfg.osiePathOverride, "osie-path-override", "", "A custom URL for OSIE/Hook images.")
 
 	return &ffcli.Command{
 		Name:       name,
@@ -415,6 +419,7 @@ func (cf *config) registerInstallers() (job.Installers, error) {
 		env.Get("REGISTRY_USERNAME"),
 		env.Get("REGISTRY_PASSWORD"),
 		env.Bool("TINKERBELL_TLS", true),
+		cf.osiePathOverride,
 	)
 	i.RegisterDistro("discovery", o.BootScript("discover"))
 	i.RegisterDefaultInstaller(o.BootScript("default"))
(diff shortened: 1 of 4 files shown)


mmlb:
Hi @ptrivedi most of this is already in progress in #258, please give that a try.

ptrivedi:
> 

We may need to discuss this. Does MIRROR_BASE_URL allow you to completely override the location URL for OSIE? Looks like it would still end up appending the 'current' directory etc. In the shorter term, this might come in handy.

installers/osie/mirror.go
```diff
@@ -42,6 +43,22 @@ func buildOSIEURL() (*url.URL, error) {
 	return u, nil
 }
 
+func getOSIEPathOverride() string {
+	base, err := url.Parse(conf.MirrorBaseUrl)
+	if err != nil {
+		panic(errors.Wrap(err, "parsing MirrorBaseUrl"))
+	}
+	if s, ok := os.LookupEnv("OSIE_PATH_OVERRIDE"); ok {
```
jacobweinstock, line 51: Hey @ptrivedi. I know looking at this file the model is to use env vars here but it's something that should be avoided. Boots' design philosophy is that 
>_"Only func main has the right to decide which flags, env variables, config files are available to the user"_ -- [ref](https://github.com/tinkerbell/boots/blob/main/docs/DESIGNPHILOSOPHY.md#functionsmethods)

This osie override option could be defined [here](https://github.com/tinkerbell/boots/blob/0199cbccfa3ad4f3bdd499e0214c04d31b822660/cmd/boots/main.go#L339) as a CLI flag. This would also automatically make an env var available. The env var would be prefaced with `BOOTS_`. So a full example would be `BOOTS_OSIE_OVERRIDE`.

mmlb:
> > 
> 
> We may need to discuss this, I'm not sure exactly this is taken care of by that PR. Does MIRROR_BASE_URL allow you to completely override the location URL for OSIE? Looks like it would still end up appending the 'current' directory etc. In the shorter term, this change/variable might come in handy because it will bypass all the code that tries to append 'misc/osie' to the osie path followed by the appending of 'current'

I'd rather do this with MIRROR_BASE_URL than add *yet another* env var/url that is hard to tell what the difference is to the others.

ptrivedi:
> 

Bigger issue is

ptrivedi:
> 

In absence of documentation, it is already hard to tell what any of these options do. But I agree, that this could be consolidated. So, if as a part of your or some future cleanup PRs, there are plans to consolidate MIRROR_BASE_URL, OSIE_PATH, MIRROR_HOST, defaultOSIEPath const, hardcoding in osieBaseURL() and workflowBaseURL() etc., I can close out this PR.

cmd/boots/main.go
```diff
@@ -86,6 +86,9 @@ type config struct {
 	kubeAPI string
 	// kubeNamespace is an override for the namespace the kubernetes client will watch.
 	kubeNamespace string
+	// osiePathOverride allows a completely custom path/URL to be specified for OSIE/Hook images
+	// This will bypass the hardcoded path appending of 'misc/osie/current' to the path
+	osiePathOverride string
```
jacobweinstock, line 91: this seems to not be used anywhere. I think it needs to be plumbed through to the `installer/osie/main.go`.

installers/osie/main.go
```diff
@@ -19,6 +22,23 @@ type installer struct {
 	// hollowParams are passed to deprovisioning instances for hardware reporting
 	// TODO(mmlb): remove this EMism now that we can use extra-kernel-args
 	hollowParams string
+	osiePathOverride string
+}
+
+func getOSIEPathOverride() string {
```
jacobweinstock, line 28: if a cli flag is to be used, i would suggest this be removed.

jacobweinstock (approved):

PR #267, opened 2022-05-07
Remove references to packet in auto.ipxe for OSIE installer:

jacobweinstock:
## Description

With the recent addition of the CLI flag `-extra-kernel-args` all of the packet (equinix metal) specific kernel args in the auto.ipxe of the OSIE installer can now be specified at runtime instead of compiled in.

## Why is this needed

Fixes: #

## How Has This Been Tested?

I have manually tested this with the sandbox repo.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/conf/config.go
+++ b/conf/config.go
@@ -30,10 +30,6 @@ var (
 
 	TrustedProxies = parseTrustedProxies()
 
-	// Hollow auth secrets, passed into osie.
-	HollowClientID            = env.Get("HOLLOW_CLIENT_ID")
-	HollowClientRequestSecret = env.Get("HOLLOW_CLIENT_REQUEST_SECRET")
-
 	// Vendor services url, used by osie to proxy requests for OS image artifacts.
 	OsieVendorServicesURL = env.Get("OSIE_VENDOR_SERVICES_URL")
 )
--- a/installers/osie/ipxe_script_test.go
+++ b/installers/osie/ipxe_script_test.go
@@ -146,12 +146,12 @@ var action2Plan2Body = map[string]map[string]string{
 
 var discoverBodies = map[string]string{
 	"c3.small.x86": `
-kernel ${base-url}/vmlinuz-${arch} ip=dhcp modules=loop,squashfs,sd-mod,usb-storage alpine_repo=${base-url}/repo-${arch}/main modloop=${base-url}/modloop-${arch} tinkerbell=${tinkerbell} syslog_host=${syslog_host} packet_action=${action} packet_state=${state} osie_vendors_url=https://localhost packet_base_url=${base-url} packet_bootdev_mac=${bootdevmac} facility=` + facility + ` intel_iommu=on iommu=pt initrd=initramfs-${arch} console=tty0 console=ttyS1,115200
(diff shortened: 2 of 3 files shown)


mikemrm (commented):
Sorry for the late response. But I think @nshalman and @mmlb need to look at this as well.

The logic for some of the removed arguments won't be able to be added with a static `extra-kernel-args`. Specifically the Hollow variables which should only be included on a deprovision.

With this change, it looks like we'll need to include in deployments the following extra arguments to match the previous arguments.

```
alpine_repo=${base-url}/repo-${arch}/main
modloop=${base-url}/modloop-${parch}
packet_action=${action}
packet_state=${state}
osie_vendors_url=$OSIE_VENDOR_SERVICES_URL
packet_bootdev_mac=${bootdevmac}
```

As well as conditionally

```
hollow_client_id=$HOLLOW_CLIENT_ID
hollow_client_request_secret=$HOLLOW_CLIENT_REQUEST_SECRET
packet_base_url=${base-url}
```

And `packet_base_url=$MIRROR_BASE_URL/workflow` when running in workflow mode.

---

I like the idea of this, but with the tinkerbell/osie being archived. Perhaps the osie boots installer package should remain mostly unchanged, and a new default installer with better configurability should be created instead? Just a thought. I'd like to see what @nshalman and @mmlb have to say.

jacobweinstock:
> Sorry for the late response. But I think @nshalman and @mmlb need to look at this as well.
> 
> The logic for some of the removed arguments won't be able to be added with a static `extra-kernel-args`. Specifically the Hollow variables which should only be included on a deprovision.
> 
> With this change, it looks like we'll need to include in deployments the following extra arguments to match the previous arguments.
> 
> ```
> alpine_repo=${base-url}/repo-${arch}/main
> modloop=${base-url}/modloop-${parch}
> packet_action=${action}
> packet_state=${state}
> osie_vendors_url=$OSIE_VENDOR_SERVICES_URL
> packet_bootdev_mac=${bootdevmac}
> ```
> 
> As well as conditionally
> 
> ```
> hollow_client_id=$HOLLOW_CLIENT_ID
> hollow_client_request_secret=$HOLLOW_CLIENT_REQUEST_SECRET
> packet_base_url=${base-url}
> ```
> 
> And `packet_base_url=$MIRROR_BASE_URL/workflow` when running in workflow mode.
> 
> I like the idea of this, but with the tinkerbell/osie being archived. Perhaps the osie boots installer package should remain mostly unchanged, and a new default installer with better configurability should be created instead? Just a thought. I'd like to see what @nshalman and @mmlb have to say.

Hey @mikemrm , thanks for the review. Providing optional auto.ipxe logic sounds like something we need to figure out and a great example of use-cases we need to solve in a generic way instead of keeping hard coded Equinix only logic like this. 

Are there certain packet/equinix references you think should be good to remove now? and any specific references that would impact the Equinix Metal specific deployment of Boots?

jacobweinstock:
> I think we at Equinix Metal need to get this unblocked. It might take us longer than ideal to fully pivot off these bits. The two main options in my mind are:
> 
> * Metal forks the repo to buy us time to do our migration while unblocking this work
> * The Tinkerbell project allows us to maintain our production branch within the project while we unblock this work and then get ourselves back onto the main branch.
> 
> ~@jacobweinstock Do you have any opinions on fork vs branch within the project?~
> 
> I've been informed that we'll be forking. We'll be out of the way shortly...

This is definitely not what I wanted to see, very unfortunate for both sides.

chrisdoherty4 (approved):

PR #269, opened 2022-05-10
kubernetes deployment manifests

jacobweinstock:
## Description

Add manifests and docs for deploying Boots to different kubernetes clusters. 

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


manifests/kustomize/base/rbac.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: tinkerbell-boots
+  namespace: tinkerbell
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
chrisdoherty4, line 8: Can this just be a `Role`?

manifests/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+   ```bash
+   # Deploy Boots to KinD
+   kubectl kustomize manifests/kustomize/overlays/kind | kubectl apply -f -
+   ```
+
+3. Watch the logs
+
+   ```bash
+   kubectl -n tinkerbell logs -f $(kubectl get -n tinkerbell pods --template '{{range .items}}{{.metadata.name}}{{end}}' --selector=app=tinkerbell-boots)
```
micahhausler, line 31: I think this (and other log watching) can be simplified to
```suggestion
   kubectl -n tinkerbell logs -f -l app=tinkerbell-boots
```

manifests/kustomize/base/rbac.yaml
micahhausler: It probably should be. The issue with boots (and hegel, and tink) is that incoming requests don't encode a namespace, so it only really makes sense to operate the components in a single namespace

manifests/kustomize/overlays/dev/nginx_ingress.yaml
```diff
@@ -0,0 +1,628 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    app.kubernetes.io/instance: ingress-nginx
+    app.kubernetes.io/name: ingress-nginx
+  name: ingress-nginx
```
micahhausler, line 7: For dev this is probably ok, but I'm a little wary of keeping ingres-nginx stuff up to date. Is there an easy, one-line way to update all this if ingress-nginx does a new release?

Tiltfile
```diff
@@ -0,0 +1,11 @@
+local_resource(
```
micahhausler, line 1: The docs don't mention tilt, can you add that if we're going to commit a Tiltfile?

manifests/README.md
```diff
@@ -0,0 +1,98 @@
...
+   ```
+
+## Kubernetes
+
+### Prerequisites
+
+This deployment requires a running Kubernetes cluster. It can be a single node cluster. It is required to be running directly on a Linux machine, not in a container.
+This deployment is under development and is not guaranteed to work at this time.
```
micahhausler, line 83: Can you mention that this also installs ingress-nginx?

micahhausler (changes requested):
This is awesome, just a few comments/questions

manifests/kustomize/overlays/dev/nginx_ingress.yaml
chrisdoherty4: I was just scouting around for the same reason. Perhaps their helm chart could be of use. 

```
kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.2.0/deploy/static/provider/cloud/deploy.yaml

https://kubernetes.github.io/ingress-nginx/deploy/#quick-start```

File: manifests/kustomize/base/deployment.yaml [hunk 6]
```diff
@@ -0,0 +1,65 @@
...
+  template:
+    metadata:
+      labels:
+        app: tinkerbell-boots
+    spec:
+      containers:
+        - args: ["-log-level", "debug", "-dhcp-addr", "0.0.0.0:67"]
+          env:
+            - name: BOOTS_EXTRA_KERNEL_ARGS
+              value: "tink_worker_image=reg.weinstocklabs.com/tink-worker:latest"
```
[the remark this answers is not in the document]
jacobweinstock: no, will update. thanks.

manifests/README.md
jacobweinstock: oh excellent, will update. thanks

manifests/kustomize/base/deployment.yaml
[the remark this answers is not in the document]
jacobweinstock: updated

manifests/README.md
```diff
@@ -0,0 +1,134 @@
...
+   # Create the K3D cluster
+   k3d cluster create --network host --no-lb --k3s-arg "--disable=traefik"
+   ```
+
+2. Deploy Boots
+
+   Start by updating `MIRROR_BASE_URL`, `PUBLIC_IP`, `PUBLIC_SYSLOG_FQDN`, and `TINKERBELL_GRPC_AUTHORITY` env variables in the `manifests/kustomize/base/deployment.yaml` file.
+
+   ```bash
+   # Deploy Boots to KinD
```
mmlb, line 72: to k3d?

rules.mk
```diff
@@ -26,7 +26,7 @@ cmd/boots/boots-linux-amd64: FLAGS=GOARCH=amd64
 cmd/boots/boots-linux-arm64: FLAGS=GOARCH=arm64
 cmd/boots/boots-linux-armv6: FLAGS=GOARCH=arm GOARM=6
 cmd/boots/boots-linux-armv7: FLAGS=GOARCH=arm GOARM=7
-cmd/boots/boots-linux-386 cmd/boots/boots-linux-amd64 cmd/boots/boots-linux-arm64 cmd/boots/boots-linux-armv6 cmd/boots/boots-linux-armv7: boots
+cmd/boots/boots-linux-386 cmd/boots/boots-linux-amd64 cmd/boots/boots-linux-arm64 cmd/boots/boots-linux-armv6 cmd/boots/boots-linux-armv7:
```
mmlb, line 29: the dep on `boots` here was to also depend on `boots`' deps, so maybe pull those out into a variable and add those as deps to these target and `boots` too?

manifests/README.md
```diff
@@ -0,0 +1,134 @@
+# Manifests
+
+This directory contains the manifests for deploying Boots to various environments.
```
mmlb, line 3: Would it makes sense to have a quick compare&contrast of these _ as they relate to boots_ (not in general). For example you mention

> **Note:** KinD will not be able to listen for DHCP broadcast traffic. Using a DHCP relay is recommended.

Does that apply to any other setups?

jacobweinstock: sure, i'll add something.

jacobweinstock: This note only applies to KinD.
Also, I've updated the readme with a visual comparison. Please check it out when you have a chance.

rules.mk
jacobweinstock: arrggg.....this was causing a double `go build` for me. Regardless, I've reverted this change. Feels out of scope for this PR anyway.

manifests/README.md
jacobweinstock: updated

manifests/README.md
mmlb: :+1: on the image (I wondered if it could be recreated in mermaidjs to avoid a binary file, but I didn't really see anything as nice as the image you have). I'm a little confused by the visualization. Does linux need both k3d and kind?

manifests/kustomize/overlays/dev/nginx_ingress.yaml
jacobweinstock: sure, updated.

manifests/README.md
jacobweinstock: I'm not quite set on using nginx as of yet. would it be alright to add this once i'm certain?

manifests/README.md
jacobweinstock: > Does linux need both k3d and kind?

No, the diagram is a comparison only.

mmlb: gotcha, maybe have them at the same "level" and avoid the arrows making it look like traffic goes from one to the other?

jacobweinstock: hmmm.....I could move the boxes around, but the arrow is coming out of the host network ns not the K3D container. the arrows purpose is to illustrate the traffic flow to the KinD container.

mmlb: I guess I'm just confused here because it looks like the image is saying we're running 2 boots containers. One in K3D and one in KinD and the one in K3D talks to KinD for some reason? I like the images, but I don't want to add any confusion if someone is looking at them. Can you do one image for K3D and one for KinD? Or some better separation so that the image isn't implying that they are communicating to each other. I think part of the confusion is also that the 2 different setups have different communication flows, why wouldn't k3D and kind both have similar flows (note I don't mean where the arrows originate/terminate but the direction K3D has outgoing arrows, KinD has incoming).

mmlb (approved):

PR #271, opened 2022-05-12
Fix regression from data-model refactor

mmlb:
## Description

An EM specific regression popped up due to #255 where `GetDiscoveryFromEM` isn't being called anymore (color me surprised thats all we noticed :smile:). This fixes it and adds some tests for the fallback behavior.

## Why is this needed

Reinstates the cacher mode behavior of falling back to trying to fetch some data from EMAPI.

## How Has This Been Tested?

Unit tests. Ran against cacher/emapi using sandbox.

No more missing fallback for EM.

### DIFF
--- a/client/reporter.go
+++ b/client/reporter.go
@@ -17,4 +17,6 @@ type Reporter interface {
 	PostInstanceFail(ctx context.Context, id string, body io.Reader) error
 	PostInstancePassword(ctx context.Context, id, pass string) error
 	UpdateInstance(ctx context.Context, id string, body io.Reader) error
+
+	Post(ctx context.Context, ref, mime string, body io.Reader, v interface{}) error
 }
--- a/cmd/boots/main.go
+++ b/cmd/boots/main.go
@@ -118,7 +118,7 @@ func main() {
 	if err != nil {
 		mainlog.Fatal(err)
 	}
-	workflowFinder, finder, err := getFinders(l, cfg)
+	workflowFinder, finder, err := getFinders(l, cfg, reporter)
 	if err != nil {
 		mainlog.Fatal(err)
 	}
@@ -221,14 +221,14 @@ func main() {
 	}
 }
 
-func getFinders(l log.Logger, c *config) (client.WorkflowFinder, client.HardwareFinder, error) {
+func getFinders(l log.Logger, c *config, reporter client.Reporter) (client.WorkflowFinder, client.HardwareFinder, error) {
 	var hf client.HardwareFinder
 	var wf client.WorkflowFinder = &client.NoOpWorkflowFinder{}
 	var err error
 
 	switch os.Getenv("DATA_MODEL_VERSION") {
 	case "":
-		hf, err = cacher.NewHardwareFinder(os.Getenv("FACILITY_CODE"))
+		hf, err = cacher.NewHardwareFinder(os.Getenv("FACILITY_CODE"), reporter)
 		if err != nil {
 			return nil, nil, err
 		}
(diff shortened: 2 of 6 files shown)


client/reporter.go
```diff
@@ -17,4 +17,6 @@ type Reporter interface {
 	PostInstanceFail(ctx context.Context, id string, body io.Reader) error
 	PostInstancePassword(ctx context.Context, id, pass string) error
 	UpdateInstance(ctx context.Context, id string, body io.Reader) error
+
+	Post(ctx context.Context, ref, mime string, body io.Reader, v interface{}) error
```
micahhausler, line 21: Does this need to be an exposed method? Can this be an internal implementation on the `cacher.HardwareFinder`? It looks like this is only called in the `cacher` package and not by external callers

micahhausler (changes requested):

client/reporter.go
mmlb: Hmm yeah thats a good point, I had this here before I figured out how to test it and thought I might need noop too. Will fix.

mmlb: hmm well actually dropping `Post` from `Reporter` sucks for other reasons. I've got 3 choices and they all have their warts.

1. Add Post to Reporter
2. Make cacher.NewHardwareFinder take a Poster and have `getFinders` type assert and bail if the Reporter isn't also a Poster
3. Leave everything as-is and make cacher.NewHardwareFinder do a runtime check and return an error.

1. is mostly unnecessary
2. we can't use Reporters everywhere
3. is going around the compiler/type system where it could help.

I was originally going to go with 3, but now that I think about it find 2 the better option. This is all terribly ugly/annoying anyway.

mmlb: I've reworked so Reporter no longer has a `Post` method. How's this look to you?

cmd/boots/main.go
```diff
@@ -221,14 +221,19 @@ func main() {
...
-func getFinders(l log.Logger, c *config) (client.WorkflowFinder, client.HardwareFinder, error) {
+func getFinders(l log.Logger, c *config, reporter client.Reporter) (client.WorkflowFinder, client.HardwareFinder, error) {
 	var hf client.HardwareFinder
 	var wf client.WorkflowFinder = &client.NoOpWorkflowFinder{}
 	var err error
 
 	switch os.Getenv("DATA_MODEL_VERSION") {
 	case "":
-		hf, err = cacher.NewHardwareFinder(os.Getenv("FACILITY_CODE"))
+		p, ok := reporter.(cacher.Poster)
```
micahhausler, line 231: Rather than expose the `Poster` interface and do this type assertion, can you just make `getDiscoveryFromEM` a private method on the `cacher.HardwareFinder`?

mmlb: the issue is that getDiscoveryEM needs a way to talk to emapi, I don't really want to duplicate bits of Reporter just for that, and I need a way to plumb down a client or creds through cacher.NewHardwareFinder so that it can be used if ByMAC needs it.

Its all ugly, we're just going back and forth on the least ugly route. I'm kinda going back to `Reporter.Post` not being that bad considering all of `Reporter` only ever really makes sense in cacher mode anyway.

micahhausler: Oh I get it, I was conflating `Finder` with `Reporter`. Then I'd probably opt for `Reporter.Post()` and avoid the type assertion

mmlb: :D yeah. So going back to v1.

micahhausler: sorry for the churn!

mmlb:
alright @micahhausler ptal again.

mmlb:
I've added the do-not-merge label because I want to get a positive run first.

micahhausler (approved):
LGTM

jacobweinstock (approved):

mmlb:
All good, verified it works as expected.

PR #281, opened 2022-06-10
Add dynamic iPXE script vars via BOOTS_IPXE_VARS env var

ScottGarman:
This allows an arbitrary number of iPXE variables to be set in all of
our iPXE scripts when they are defined via the BOOTS_IPXE_VARS env var.
The env var should contain space-separated variable definitions, e.g.
'myvar1=myval1 myvar2=myval2'.


conf/config.go
```diff
@@ -36,8 +36,41 @@ var (
 
 	// Vendor services url, used by osie to proxy requests for OS image artifacts
 	OsieVendorServicesURL = env.Get("OSIE_VENDOR_SERVICES_URL")
+
+	// BOOTS_IPXE_VARS can optionally be used to inject any number of variables
+	// into the iPXE script
+	DynamicIPXEVars = parseDynamicIPXEVars(env.Get("BOOTS_IPXE_VARS"))
```
jacobweinstock, line 42: We're trying to move away from the pattern this file follows of calling environment variables. 

_"Only func main has the right to decide which flags, env variables, config files are available to the user "_ -- [ref](https://github.com/tinkerbell/boots/blob/main/docs/DESIGNPHILOSOPHY.md)

Boots' design philosophy points env var to func main. Ideally, i think we would want to have this be a CLI flag/env combo and then plumbed through to these installers. See this [func](https://github.com/tinkerbell/boots/blob/51fec55b27aa94deda3367e717bb19016d7eec0c/cmd/boots/main.go#L366) for the others using this pattern.

jacobweinstock (commented):
Hey @ScottGarman, thanks for this PR. I'm curious what use-case(s) you're interested in solving with this functionality? 

Arbitrary iPXE variables on their own don't seem particularly useful.  Is this so that existing iPXE variables can be overridden? something else, possibly?

conf/config.go
ScottGarman: Hey Jacob! We have an interesting issue where we're downloading boot assets via an https url, but the host making the request may have a networking configuration restricted to Layer 2, and thus be unable to validate the tls cert from ca.ixpe.org. We do have a local proxy service available, but to use it we need to set the iPXE var `crosscert`.

Originally the idea was to add yet another env var to boots, which we know is not optimal. But with this approach, at least we can add only one env var and potentially solve this class of problem again without touching more boots code. If `BOOTS_IPXE_VARS` isn't set, then the ipxe scripts will remain the same as before. But now we can do things in the helm chart such as:

```
env:
  - name: BOOTS_IPXE_VARS
    value: |
      vendors_domain=vendors.{{ .clusterInfo.fqdn }}
      crosscert=http://${vendors_domain}/ipxe/ca
```

That's the background on the motivation behind this. Let me see what I can do to make this a CLI flag according to your suggestion.

jacobweinstock: right on, thank you for sharing that! very helpful. I had forgotten that `set` in ipxe does more than just arbitrary variable setting.

Would you also be open to adding a code comment with a link to the ipxe docs around variables like this? https://ipxe.org/cfg

jacobweinstock:
Just food for thought, there is at least one other way to accomplish the `crosscert` setting. DHCP option 175 can be used to set ipxe specific config values. [ref](https://ipxe.org/cfg/crosscert)

Boots already has this available. Something like this: `OptionCrossCert: []byte("http://ca.ipxe.org/auto.custom"),` could be added here:
https://github.com/tinkerbell/boots/blob/51fec55b27aa94deda3367e717bb19016d7eec0c/ipxe/dhcp_options.go#L63

ScottGarman:
@jacobweinstock I decided not to go down the route of using DHCP options for this, but I believe I've incorporated the rest of your feedback. Please re-review.

cmd/boots/main.go
```diff
@@ -68,8 +68,12 @@ type config struct {
 	ipxeHTTPEnabled bool
 	// ipxeRemoteTFTPAddr is the address of the remote TFTP server serving iPXE binaries
 	ipxeRemoteTFTPAddr string
-	// remoteTFTPPort is the address of the remote HTTP server serving iPXE binaries
+	// ixpeRemoteHTTPAddr is the address and port of the remote HTTP server serving iPXE binaries
```
jacobweinstock, line 71: was this intentional?

cmd/boots/main.go
```diff
@@ -294,6 +298,33 @@ func getReporter(l log.Logger) (client.Reporter, error) {
...
+	retVal := make([][]string, len(allDefs))
+
+	// iterate over each variable definition and split on '='
+	for i, el := range allDefs {
+		varDef := strings.SplitN(el, "=", 2)
+		if len(varDef) == 2 {
+			retVal[i] = varDef
+		} else {
+			err := errors.Errorf("unable to parse IPXE dynamic var definitions from string %v", v)
+			panic(err)
```
jacobweinstock, line 321: What do you think about returning an error from this function and letting that propagate up so that `func main` can handle the panic?

installers/custom_ipxe/main.go
```diff
@@ -45,6 +52,11 @@ func bootScript(ctx context.Context, j job.Job, s *ipxe.Script) {
 		return
 	}
 
+	if i.extraIPXEVars != nil {
```
jacobweinstock, line 55: I don't this this check is necessary. I believe an empty array/nil value here will have nothing to iterate and will just move on. [example](https://goplay.tools/snippet/VOVP-oixwnI).

installers/osie/main.go
```diff
@@ -122,6 +130,12 @@ func (i installer) rescue(ctx context.Context, j job.Job, s *ipxe.Script) {
 }
 
 func (i installer) discover(ctx context.Context, j job.Job, s *ipxe.Script) {
+	if i.extraIPXEVars != nil {
+		for index := 0; index < len(i.extraIPXEVars); index++ {
```
jacobweinstock, line 134: any reason this loop is different from the others?

jacobweinstock (changes requested):

cmd/boots/main.go
ScottGarman: Yeah, that was fixing a comment that was incorrect.

jacobweinstock: oh doh, for some reason i did even see the `//`. sorry. Does look like the word is misspelled?

installers/vmware/main.go
```diff
@@ -46,11 +53,17 @@ func (i installer) BootScript(slug string) job.BootScript {
...
 
 	return func(ctx context.Context, j job.Job, s *ipxe.Script) {
-		script(j, s, path)
+		script(i, j, s, path)
 	}
 }
 
-func script(j job.Job, s *ipxe.Script, basePath string) {
+func script(i installer, j job.Job, s *ipxe.Script, basePath string) {
+	if i.extraIPXEVars != nil {
```
displague, line 61: [same comment](https://github.com/tinkerbell/boots/pull/281/files/344c786c99e2ece0c33702be22e98f6f48e9ff65#r896185299) about this loop

cmd/boots/main.go
ScottGarman: Fixed.

cmd/boots/main.go
ScottGarman: Yeah, that seems reasonable. This should now be implemented.

installers/custom_ipxe/main.go
ScottGarman: Good catch, thank you!

installers/vmware/main.go
ScottGarman: Good catch, thank you!

installers/osie/main.go
ScottGarman: This was my original approach before I learned about using range for this, I had just missed updating it. This should now be fixed.

cmd/boots/main.go
```diff
@@ -294,6 +298,34 @@ func getReporter(l log.Logger) (client.Reporter, error) {
...
+// string, and return a an array of two-element arrays which are the key/value
+// string pairs of the variable's name and value. These will later be injected
+// as variable definitions in the iPXE script. Variable order is preserved.
+func parseDynamicIPXEVars(v string) ([][]string, error) {
+	if v == "" {
+		return nil, nil
+	}
+
+	// extract multiple variable definitions by splitting first on spaces
+	allDefs := strings.Fields(v)
```
displague, line 311: I'm wondering if we might run into problems with spaces within a field, where a quoted field value would be needed. 

https://ipxe.org/cfg -- scriptlet for example (https://forum.ipxe.org/printthread.php?tid=11058 has some examples)

https://stackoverflow.com/a/47489825/156674 has some suggestions over strings.Fields (not all of which apply in this situation, but offer some thought)

ScottGarman: Sounds like I could really use a unit test for this function and include some of these edge cases. I'll add that next.

cmd/boots/main.go
```diff
@@ -294,6 +298,34 @@ func getReporter(l log.Logger) (client.Reporter, error) {
 	}
 }
 
+// parseDynamicIPXEVars will parse any number of variable definitions from a
+// string, and return a an array of two-element arrays which are the key/value
+// string pairs of the variable's name and value. These will later be injected
+// as variable definitions in the iPXE script. Variable order is preserved.
+func parseDynamicIPXEVars(v string) ([][]string, error) {
```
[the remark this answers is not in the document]
ScottGarman: Yes, I know have a unit test for this in main_test.go.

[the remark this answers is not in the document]
ScottGarman: Good catch. I added a test case for this in the unit tests and fixed things so we do return an error in the case that either the key/value are empty strings.

cmd/boots/main_test.go
```diff
@@ -45,6 +46,90 @@ func TestParser(t *testing.T) {
...
+					return
+				}
+				t.Fatalf("parseDynamicIPXEVars() did not return an error, instead returned %v", got)
+			}
+			if err != nil {
+				t.Fatalf("parseDynamicIPXEVars() returned an unexpected error: %s", err)
+			}
+
+			want := tt.want
+			if !reflect.DeepEqual(want, got) {
```
jacobweinstock, line 126: You might want to use `cmp.Diff` or `cmp.Equal` instead of `reflect.DeepEqual`.

_"You will find older code using the standard reflect.DeepEqual function to compare complex structures. Prefer cmp for new code, and consider updating older code to use cmp where practical. reflect.DeepEqual is sensitive to changes in unexported fields and other implementation details."_ -- [ref](https://github.com/golang/go/wiki/TestComments#equality-comparison-and-diffs)

jacobweinstock (changes requested):
Thanks @ScottGarman for your patience with this review. It's looking good. I just have hopefully one quick suggestion.

jacobweinstock (approved):
Thanks Scott!

cmd/boots/main_test.go
ScottGarman: Cool, I've now switched this to using cmp.Equal.

PR #282, opened 2022-06-30
Introduction of report for k8s client

ibrokethecloud:
## Why is this needed

Currently the k8s reporter is not implemented, as a result among other things the DisablePXE method is a no-op.

As a result in scenarios when leveraging custom-pxe, no DisablePXE is performed.

The PR introduces a reporter implementation of the k8s client.  This allows the implementation for DisablePXE implementatino.

In addition, the OSIE / boots events are now also logged as k8s events associated with the hardware objects.

Fixes: #

## How Has This Been Tested?

A simple unit test leveraging envTest is available which can simulate the events and hardware updates.

## Checklist:

I have:

- [ ] added unit or e2e tests


jacobweinstock:
The code base has changed quite a bit since this went stale. This reporter/phone home logic has mostly been deprecated. As this would add webhook functionality to Boots and require write permissions in kubernetes to hardware spec, i think this would need to be discussed a little bit more. Feel free to reopen and start the conversation. Either way, thanks for using and contributing!

PR #283, opened 2022-07-01
Fix and enforce linter

jacobweinstock:
## Description

This gets the repo to a baseline of linting based on the .golangci.yml file generated from running [lint-install](https://github.com/tinkerbell/lint-install).

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


client/kubernetes/hardware_finder.go
```diff
@@ -123,14 +123,15 @@ func (f *Finder) HasActiveWorkflow(ctx context.Context, hwID client.HardwareID)
 
 	wfContexts := []*workflow.WorkflowContext{}
 	for _, wf := range stored.Items {
+		wf := wf
```
mmlb, line 126: this was definitely buggy before

cmd/boots/dhcp_test.go
```diff
@@ -68,5 +66,5 @@ func TestMain(m *testing.M) {
 	defer l.Close()
 	mainlog = l.Package("main")
 	metrics.Init(l)
-	os.Exit(m.Run())
+	os.Exit(m.Run()) //nolint:gocritic // this seems to be the correct pattern
```
mmlb, line 69: yep

dhcp/dhcp.go
```diff
@@ -11,7 +11,7 @@ type Reply interface {
 }
 
 func NewReply(w dhcp4.ReplyWriter, req *dhcp4.Packet) Reply {
-	switch req.GetMessageType() {
+	switch req.GetMessageType() { //nolint:exhaustive // Boots only supports responding to these message types
```
mmlb, line 14: you could also do a `default:` case that doesn't do anything. This way one less `//nolint`.

mmlb (commented):
fantastic!. A couple of minor comments and one nit. The nit is that you're mixing 2 different ways to define functions whose arguments are all ignored. Some changes you're doing here are like:

```go
func(_ type1, _ type2, _ type3)
```
and
```go
func(type1, type2, type3)
```
I prefer the latter and I think its the most common way. Can you make it so only one way is used please?

dhcp/dhcp.go
jacobweinstock: makes sense. updated.

jacobweinstock:
> fantastic!. A couple of minor comments and one nit. The nit is that you're mixing 2 different ways to define functions whose arguments are all ignored. Some changes you're doing here are like:
> 
> ```go
> func(_ type1, _ type2, _ type3)
> ```
> 
> and
> 
> ```go
> func(type1, type2, type3)
> ```
> 
> I prefer the latter and I think its the most common way. Can you make it so only one way is used please?

yeah, seems reasonable. updated.

mmlb:
:( I always forget to check of the "viewed" button while reviewing and then need to go through the whole thing again later ...

installers/customipxe/errors.go
```diff
@@ -0,0 +1,5 @@
+package customipxe
+
+import "errors"
+
+var ErrEmptyIpxeConfig = errors.New("ipxe config URL or Script must be defined")
```
mmlb, line 5: Another nit, one that wasn't caught by gometalinter but is in the spirit of one of its linters: This should be `ErrEmptyIPXEConfig` since its `iPXE` because PXE is an acronym like DNS, ID, IP, ...  and should be all caps. The `i` should also be caps because its a new word. Not a blocker if you just want to ignore though.

mmlb (commented):
just one thing I noticed now

installers/customipxe/errors.go
jacobweinstock: yeah, that's interesting that the linter didn't catch that. I've updated it regardless.

mmlb: ha!

mmlb: wait what... that was way too fast. I'm pretty sure github sent that to you before I submitted it because it complained about needing a comment for the review as a whole...

mmlb (approved):

PR #285, opened 2022-08-08
support esxi 7.0.1+ root password hash changes

mikemrm:
## Description

The password on an instance is typically a generated password.
For VMWare an alternative password hash can be provided via the
instance's customdata field.

Previously the method for utilizing that override required manipulating the shadow file.
This method is no longer supported by vmware, as the shadow file is generated based off a database value.

The correct way to set the root password is through the kickstart file.
We already were setting the generated password, so we could simply add
the necessary logic to check for an overwritten password instead.

## Why is this needed

The previous hack of editing /etc/shadow no longer works as that file is overwritten at startup.

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/client/instance.go
+++ b/client/instance.go
@@ -33,6 +33,10 @@ type Instance struct {
 	// Only returned in the first 24 hours of a provision
 	PasswordHash string `json:"password_hash,omitempty"`
 
+	// CustomData is an additional JSON field included in metadata
+	// Typically in the format of map[string]interface{}, however this format is not required.
+	CustomData interface{} `json:"customdata,omitempty"`
+
 	Tags []string `json:"tags,omitempty"`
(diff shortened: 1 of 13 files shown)


ScottGarman (approved):
Looks good, pending testing.

mikemrm:
Whoops, forgot to note, this has been tested on both 6.7 and 7.0 as well as 7.0.3 (the one that initially gave us trouble) on multiple plans, all were good!

Thanks!

displague (approved):

ScottGarman (approved):

nshalman (approved):

nshalman:
@Mergifyio refresh

PR #288, opened 2022-08-30
unify AllowPXE so checks are consistent

mikemrm:
## Description

isPXEAllowed is checked in most locations internally to determine if
dhcp should hand off the ipxe image etc.

isPXEAllowed checks both the hardware and instance for whether it's
allowed to pxe.

AllowPXE on the other hand is used only for returning the auto.ipxe
script and only checks if the hardware is allowed to pxe.

This inconsistency allows dhcp to provide the ipxe image but then the
ipxe image fails to load the script if the instance AllowPXE is true
while hardware AllowPXE is false.

This change makes the AllowPXE check consistent everywhere.

## Why is this needed

Simplifies PXE checks

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


mmlb (approved):
lgtm, thanks!

PR #291, opened 2022-09-21
installers(vmware): improve bootdrivehint matching

dustinmiller:
## Description

Influences the installation drive for these plans in a beneficial way, the smaller one

## Why is this needed

Client preferences

## How Has This Been Tested?

With love and CI/CD


nuclearbob:
I don't know this codebase well, but the changes make sense to me.

jacobweinstock (approved):

mmlb (approved):

PR #292, opened 2022-09-21
Optionally add vlan id to DHCP option 43.116

jacobweinstock:
## Description

This enables Boots to setup DHCP option 43, sub option 116 with the VLAN ID pulled from the hardware record. This also adds a kernel command line parameter to the default (OSIE) installer auto.ipxe script with the vlan id (`vlan_id=`). This will enable Hook to create a tagged vlan interface when it boots up. This is all inline with proposal [0030](https://github.com/tinkerbell/proposals/pull/60). 

This is only being implemented with the Kubernetes backend. The Tink+Postgresql is [deprecated](https://github.com/tinkerbell/proposals/tree/main/proposals/0029) and the Cacher backend is not officially deprecated (yet) but should be considered legacy. 

## Why is this needed

Fixes: #

## How Has This Been Tested?

Manual testing following the K3D guide [here](https://github.com/tinkerbell/boots/blob/main/docs/manifests/k3d.md).

No impacts.

## Checklist:

I have:

- [ ] added unit or e2e tests


dhcp/config.go
```diff
@@ -118,3 +118,26 @@ func (c *Config) SetDNSServers(ips []net.IP) {
...
+		c := make(dhcp4.OptionMap, 0)
+		cur = c.Serialize()
+	}
+	// deserialize it into a dhcp4.OptionMap
+	n := make(dhcp4.OptionMap, 0)
+	if err := n.Deserialize(cur, nil); err != nil {
+		dhcplog.Info("unable to deserialize option 43")
+
+		return
+	}
```
mmlb, line 139: seems weird to do this check even when we know there was no previous Opt43. Why not move the Deserialize into an `else` branch of the above `if`? Actually I don't see the poing of doing anything at all when there is no pre-existing opt43 set, doesn't this work the same with less code?:

```go
n := make(dhcp4.OptionMap, 0)
if cur, ok := c.opts.GetOption(dhcp4.OptionVendorSpecific); ok {
	if err := n.Deserialize(cur, nil); err != nil {
		dhcplog.Info("unable to deserialize option 43")

		return
	}
}
```

*Actually*, after messing with similar code in `dhcp/pxe.go` why not just directly use the `OptionMap`?:

```go
n := make(dhcp4.OptionMap, 0)

// get current option 43
if cur, ok := c.opts.GetOption(dhcp4.OptionVendorSpecific); ok {
	if err := n.Deserialize(cur, nil); err != nil {
		dhcplog.Info("unable to deserialize option 43")

		return
	}
}

n[subOpt] = []byte(s)
```

dhcp/pxe.go
```diff
@@ -126,13 +126,24 @@ func SetupPXE(ctx context.Context, rep, req *dhcp4.Packet) bool {
...
+	err := pxeVendorOptions.Deserialize(cur, nil) // merge the current option 43 with pxeVendorOptions.
+	if err != nil {
+		pxeVendorOptions = dhcp4.OptionMap{
+			6:  []byte{0x8}, // PXE_DISCOVERY_CONTROL: Attempt to tell PXE to boot faster.
+			69: binaryTpFromContext(ctx),
+		}
+	}
 
-	rep.SetOption(dhcp4.OptionVendorSpecific, pxeVendorOptions)
+	rep.SetOption(dhcp4.OptionVendorSpecific, pxeVendorOptions.Serialize())
```
mmlb, line 146: similarly this seems like less/easier-to-follow code:
```go
om := make(dhcp4.OptionMap, 0)

cur, ok := rep.GetOption(dhcp4.OptionVendorSpecific)
if ok {
	om.Deserialize(cur, nil)
}

om[6] = []byte{0x8} // PXE_DISCOVERY_CONTROL: Attempt to tell PXE to boot faster.
om[69] = binaryTpFromContext(ctx)

rep.SetOption(dhcp4.OptionVendorSpecific, om.Serialize())
```

Its easier to work with `OptionMap`s (plain old go maps) than `[]byte`s through the `Serialize`/`Deserialize` methods imo.

mmlb (commented):
couple of comments with what I think are easier to read code. It'd be nice if you could squash the `WIP` and `Set DHCP option 43.116 with VLAN ID` commits into just one.

mmlb:
One more thing (which is a minor nit) but `foo := make(somemap, 0)` seems like premature optimization/no very idiomatic vs `foo := make(somemap)` (tbh I prefer `foo := dhcp4.OptionMap{}` whenever possible). `dhcp4.OptionMap.Serialize` seems to dothe right thing and just iterate over that map, so its not like it encodes extra space go may allocate.

mmlb (approved):

PR #298, opened 2022-11-06
Add a namespace for the cluster client

moadqassem:
## Description
Currently, the kube client for boots needs a cluster role permission to fetch Hardware and Workflow CRs, even if the resources and the client are namespaced. This will lead to extending boots rbac to cluster role instead of using namespaced roles. It seems that, the current client creation doesn't scope the client to a namespace(the rest config type that is returned doesn't have a context, thus there will be no Namespace scoped anywhere in the client).     

## Why is this needed
Enabling kube client operation on Namespaced scoped resources

Fixes: #

## How Has This Been Tested?

Ran against a live cluster and the indexer succeeded to list the needed resources. 

They are not as they can pass whatever rbacs they like, based on their setup. 

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/client/kubernetes/hardware_finder.go
+++ b/client/kubernetes/hardware_finder.go
@@ -46,7 +46,7 @@ func NewFinder(logger log.Logger, k8sAPI, kubeconfig, kubeNamespace string) (*Fi
 		return nil, err
 	}
 
-	cluster, err := NewCluster(config)
+	cluster, err := NewCluster(config, kubeNamespace)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
--- a/client/kubernetes/hardware_finder.go
+++ b/client/kubernetes/hardware_finder.go
@@ -27,7 +27,7 @@ type Finder struct {
 //
 // Callers must instantiate the client-side cache by calling Start() before use.
 func NewFinder(logger log.Logger, k8sAPI, kubeconfig, kubeNamespace string) (*Finder, error) {
-	// TODO(moadqassem): Maybe use the tinkerbell kubecleint instead of using this cluster client similar to hegel.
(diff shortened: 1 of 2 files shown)


client/kubernetes/hardware_finder.go
```diff
@@ -46,7 +47,7 @@ func NewFinder(logger log.Logger, k8sAPI, kubeconfig, kubeNamespace string) (*Fi
 		return nil, err
 	}
 
-	cluster, err := NewCluster(config)
+	cluster, err := NewCluster(config, kubeNamespace)
```
chrisdoherty4, line 50:
```suggestion
	cluster, err := NewCluster(config, ccfg.Namespace())
```

This suggestion doesn't make any behavioral difference but its seemingly more resilient to config loading rules being tweaked in the lines above. It ensures we always pluck the namespace the config loader deduced irrespective of how it got it.

We could, alternatively, change `NewCluster` to just accept `ccfg` and have it pull the information from a single data structure (that honestly seems preferable if you're willing).

client/kubernetes/hardware_finder.go
```diff
@@ -27,6 +27,7 @@ type Finder struct {
 //
 // Callers must instantiate the client-side cache by calling Start() before use.
 func NewFinder(logger log.Logger, k8sAPI, kubeconfig, kubeNamespace string) (*Finder, error) {
+	// TODO(moadqassem): Maybe use the tinkerbell kubecleint instead of using this cluster client similar to hegel.
```
displague, line 30:
```suggestion
	// TODO(moadqassem): Maybe use the tinkerbell kubeclient instead of using this cluster client similar to hegel.
```

chrisdoherty4 (approved):

displague (approved):

PR #315, opened 2023-02-02
Refactor auto.ipxe and custom ipxe scripting:

jacobweinstock:
## Description

The old "installers" package had quite a few iPXE scripts to build for different operating systems. Now that only an "installer" for Hook is needed and for chain loading a script or providing an arbitrary script, we can make building the auto.ipxe script for Hook simpler and easier to understand. This refactor seeks to make it more clear for developers what the `auto.ipxe` script will be. Same for what was previously called custom_ipxe. No longer is it needed to specify the operating system to be "custom_ipxe". If
`hardware.spec.interfaces[0].netboot.ipxe.url` or
`hardware.spec.interfaces[0].netboot.ipxe.contents` is specified then it is used instead of the default `auto.ipxe` for Hook.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


ipxe/custom.go
```diff
@@ -0,0 +1,23 @@
+package ipxe
+
+import "net/url"
+
+// CustomScript is the template for the custom script.
+// It will either chain to a URL or execute an iPXE script.
+var CustomScript = `#!ipxe
+
+echo Loading the Tinkerbell iPXE custom script...
```
chrisdoherty4, line 9: Would this read better as the following?

```suggestion
echo Loading custom Tinkerbell iPXE script ...
```

ipxe/ipxe.go
```diff
@@ -0,0 +1,20 @@
+package ipxe
+
+import (
+	"bytes"
+	"html/template"
+)
+
+func GenerateTemplate[data any](d data, script string) (string, error) {
```
chrisdoherty4, line 8: This shouldn't need to be generic because `t.Execute() accepts the `any` type.


```suggestion
func GenerateTemplate(d any, script string) (string, error) {
```

jacobweinstock: ah thanks. updated

ipxe/custom.go
jacobweinstock: sure, updated.

chrisdoherty4 (approved):

PR #318, opened 2023-02-10
Expose ipxedust script patching feature

fintelia:
## Description

Update ipxedust and exposes the binary patching functionality added in https://github.com/tinkerbell/ipxedust/pull/63

## Why is this needed

New feature

N/A

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/boots/main.go
+++ b/cmd/boots/main.go
@@ -88,6 +88,8 @@ type config struct {
 	// osiePathOverride allows a completely custom path/URL to be specified for OSIE/Hook images
 	// This will bypass the hardcoded path appending of 'misc/osie/current' to the path
 	osiePathOverride string
+	// iPXE script fragment to patch into binaries served over TFTP and HTTP
+	ipxeScriptPatch string
 }
 
 func main() {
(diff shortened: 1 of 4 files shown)


conf/config.go
```diff
@@ -21,6 +21,8 @@ var (
 	HTTPBind   = env.Get("HTTP_BIND", PublicIPv4.String()+":80")
 	BOOTPBind  = env.Get("BOOTP_BIND", PublicIPv4.String()+":67")
 
+	ScriptPatch = env.Get("IPXE_SCRIPT_PATCH", "")
```
jacobweinstock, line 24: This is an old pattern that we're trying to move away from. _"Only func main has the right to decide which flags, env variables, config files are available to the user"_ -- [ref](https://github.com/tinkerbell/boots/blob/main/docs/DESIGNPHILOSOPHY.md#functionsmethods)

Would you mind adding this as a cli flag [here](https://github.com/tinkerbell/boots/blob/c92674c5f39f7005602bc6882afd035dc0f13d0c/cmd/boots/main.go#L359) instead? This will also make the cli flag name available as an env var (`BOOTS_IPXE_SCRIPT_PATCH`).

jacobweinstock (changes requested):
Thanks @fintelia! One change requested, please.

Issue #329, opened 2023-05-10
Customscript error handling does not seem correct


rgl:
The code at https://github.com/tinkerbell/boots/blob/384b68b37cdf2861158ed756ab102ab13ab1971d/ipxe/auto.go#L85-L89 does not seem correct.

When, for some reason, there is a problem interpreting what we put in the Hardware object (e.g. an invalid url), the error is never sent/displayed anywhere, and it will make things really hard to troubleshoot.

And, IMHO, the worse part, it will just boot the default iPXE script, which is not what I intent to do.

This should be changed to not ignore any errors (and not having a custom script in the Hardware object should not be considered an error too). When there is an error, it should log the error and abort the boot.

Also, returning a 404 when there is error at https://github.com/tinkerbell/boots/blob/384b68b37cdf2861158ed756ab102ab13ab1971d/ipxe/auto.go#L91-L92 does not seem correct too and should be changed to a 500.

What do you think?

jacobweinstock:
Hey @rgl, thanks for the report. Yeah, that all sounds very reasonable. Would you like to work on this? Totally fine if not. I will  take it on, no problem. I'm fine either way, let me know. thanks again for the report!

jacobweinstock:
Hey @rgl , i put a PR together for this. Still want to add a few more tests. Please check it out and let me know what you think. Thanks!

https://github.com/tinkerbell/boots/pull/330

rgl:
@jacobweinstock, its looking good.

But now that I've read a bit more about the code, I think, I will be stuck. As there is no way for a custom script/custom url to have the data about the Hardware and Tinkerbell in the Go template. So the custom boot script will not known how to reach Tinkerbell/Hegel to have the Hardware details.

I have to say that I'm trying to replace Hook OSIE with [Debian OSIE](https://github.com/rgl/tinkerbell-debian-osie). But for that, in the custom script, I think I need to be able to have access to the same data as hook in https://github.com/tinkerbell/boots/blob/384b68b37cdf2861158ed756ab102ab13ab1971d/ipxe/auto.go#L118-L156

rgl:
I'm also trying to play with [ORAS](https://github.com/oras-project/oras) using the [zot registry](https://github.com/project-zot/zot), which means the files will be served from an OCI registry, which means the urls will be somewhat odd like this:

```
[...]

PR #330, opened 2023-05-11
Fix auto and custom ipxe script generation

jacobweinstock:
## Description

Fixes for auto and custom ipxe script generation. Handle errors messages from customScript func. Return 500 when there is an error generating a script. Don't fall through to the default Hook script if a custom script exists but fails to generate properly.

## Why is this needed

Fixes: #329 

## Checklist:

I have:

- [ ] added unit or e2e tests


ipxe/auto.go
```diff
@@ -140,6 +171,9 @@ func (h *Handler) defaultScript(span trace.Span, hw client.Discoverer, ip string
 func (h *Handler) customScript(hw client.Discoverer, ip string) (string, error) {
 	mac := hw.GetMAC(net.ParseIP(ip))
 	if chain := hw.Hardware().IPXEURL(mac); chain != "" {
+		if !strings.HasPrefix(chain, "http") && !strings.HasPrefix(chain, "https") {
```
rgl, line 174: instead of checking it here, can this be done after `url.Parse` error checking and test against `u.Scheme`? that way, the dirty business of parsing urls is left to the url library.

ipxe/auto.go
```diff
@@ -76,27 +78,46 @@ func (h *Handler) HandlerFunc() http.HandlerFunc {
...
-		if cs, err := h.customScript(hw, ip); err == nil {
-			script = []byte(cs)
-			break
-		}
 		s, err := h.defaultScript(span, hw, ip)
 		if err != nil {
-			w.WriteHeader(http.StatusNotFound)
-			err := errors.Errorf("boot script %q not found", name)
+			w.WriteHeader(http.StatusInternalServerError)
+			err := errors.Wrap(err, fmt.Sprintf("boot script %q not found", name))
```
rgl, line 103: maybe change this error message to be alike the custom one? e.g. `error with default ipxe script`

rgl (commented):
looking good! thank you!

ipxe/auto.go
jacobweinstock: sure thing, will update.

ipxe/auto.go
jacobweinstock: sure thing, will update

rgl (dismissed):

PR #336, opened 2023-08-03
Remove kernel parameter ip=dhcp:

jacobweinstock:
## Description

When a machine had multiple interfaces that weren't all plugged in, this would cause significant delay in booting into Hook.

If necessary, this value can be added back in via the `-extra-kernel-args` cli flag.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


jacobweinstock:
duplicate of https://github.com/tinkerbell/boots/pull/339 and #339 doesn't have an merge conflict issues.

PR #364, opened 2023-11-03
bring back the code from tinkerbell/dhcp

jacobweinstock:
## Description

The DHCP library was created in order to provide a mechanism for the DHCP functionality to be written in a very clean and composible way. This was not possible with the previous state of Boots and the personnel involved at that time.

The current state of Smee is ready for this functionality to be returned and placed in a very clean and composable way into the code base.

Moving this functionality back in will improve the development process/experience. It is important to note that we do take on the trade off of not having the combined commit history of the DHCP library within Smee. We won't lose the commit history entirely as the DHCP library GitHub repo will remain (archived) in order to have the commit history available.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


chrisdoherty4:
Given we aren't in the business of maintaining libraries, would you be open to nesting under a top level `internal` package?

PR #434, opened 2024-04-22
smee: toDHCPData() from CRD's TimeServers to DHCP NTPServers

rpardini:
## Description

smee: toDHCPData() from CRD's TimeServers to DHCP NTPServers


jacobweinstock (approved):
Thanks, @rpardini

PR #493, opened 2024-08-05
detectPublicIPv4: try to detect ipv4's from interfaces with a default gateway first

rpardini:
#### detectPublicIPv4: try to detect ipv4's from interfaces with a default gateway first

- adds github.com/vishvananda/netlink to get at the routing tables
- if none found, old logic is kept (gets the first IPv4 from the first interface that has one)

Signed-off-by: Ricardo Pardini <ricardo@pardini.net>


rpardini:
Yay got it green, finally. Sorry for the delay...

jacobweinstock (approved):

PR #543, opened 2024-11-05
Serve and patch Hook as an ISO during runtime

rahulbabu95:
## Description
This change adds the ability in Smee to serve Hook as an ISO file through HTTP and also patch it with all the necessary dynamic information needed to connect with a Tink Server. 

## Why is this needed
Hook recently added an [feature](https://github.com/tinkerbell/hook/pull/241) to support building it as an ISO file. This change leverages that feature to be able to serve and patch Hook as an ISO file. This will be needed to be able to get Tinkerbell stack running on L3 without any Layer 2 dependency.  
 

## How Has This Been Tested?
Built a custom smee image with above changes, and tested with that image and replacing the smee deployment of Tink stack on a k3d cluster with real hardware. Set the Hardware to boot from Virtual media mount and verified that the hardware was able to boot properly.

ISO serving will not be enabled by default in the CLI and users have to explicitly set the 'iso-enabled' flag to make use of the feature. So as such the change will not have any impact on existing users.

### DIFF
--- a/internal/iso/iso.go
+++ b/internal/iso/iso.go
@@ -1,0 +1,226 @@
+package iso
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"path"
+	"path/filepath"
+	"strings"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"github.com/go-logr/logr"
+	"github.com/tinkerbell/smee/internal/dhcp/handler"
+)
+
+const (
+	defaultConsoles = "console=ttyS1 console=ttyS1 console=ttyS0 console=ttyAMA0 console=ttyS1 console=tty0"
+)
+
+type Handler struct {
+	Logger  logr.Logger
+	Backend handler.BackendReader
+	// SourceISO is the source url where the unmodified iso lives
+	// patch this at runtime, should be a HTTP(S) url.
+	SourceISO          string
+	ExtraKernelParams  []string
+	Syslog             string
+	TinkServerTLS      bool
+	TinkServerGRPCAddr string
+	// parsedURL derives a url.URL from the SourceISO
+	// It helps accessing different parts of URL
+	parsedURL *url.URL
+	// MagicString is the string pattern that will be matched
+	// in the source iso before patching. The field can be set
+	// during build time by setting this field.
+	// Ref: https://github.com/tinkerbell/hook/blob/main/linuxkit-templates/hook.template.yaml
+	MagicString string
+}
+
+func (h *Handler) RoundTrip(req *http.Request) (*http.Response, error) {
+	h.Logger.V(1).Info("entered the roundtrip func")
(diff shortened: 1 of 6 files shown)


internal/iso/iso.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	// Reverse Proxy modifies the request url to
+	// the same path it received the incoming request.
+	// Manually override the path as the incoming request has mac address appended to the req path
+	// Ex: requesting <ip-address>:port/<mac-id>/hook.iso on the Source server would error out
+	req.URL.Path = h.parsedURL.Path
+	resp, err := h.RoundTripper.RoundTrip(req)
+	if err != nil {
+		h.Logger.Info("HTTP request didn't receive a response", "error", err)
```
jacobweinstock, line 97: HTTP request to the `SourceISO` URL in this case. Might want to add the `SourceISO` to the keysAndValues and make the message part clear that this is the case too.

internal/iso/iso.go
```diff
@@ -0,0 +1,188 @@
...
+		return nil, err
+	}
+	resp.Header.Set("Server", "Tinkerbell")
```
jacobweinstock, line 100: Why do we need this?

internal/iso/iso.go
```diff
@@ -0,0 +1,188 @@
...
+	switch req.Method {
+	case http.MethodHead:
+		h.Logger.Info("HTTP HEAD request received")
```
jacobweinstock, line 103: what value does this log line provide? especially at the default level.

internal/iso/iso.go
```diff
@@ -0,0 +1,188 @@
...
+		if err != nil {
+			h.Logger.Error(err, "reading response bytes", "response", resp.Body)
+			return notFoundResponse, nil
+		}
+		if err := resp.Body.Close(); err != nil {
+			h.Logger.Error(err, "closing response body", "response", resp.Body)
+			return notFoundResponse, nil
+		}
+
+		// TODO: revisit later to handle the magic string being spread across two chunks.
```
jacobweinstock, line 120: might be helpful to elaborate on this problem here in the comment.

jacobweinstock (changes requested):

internal/iso/iso.go
```diff
@@ -0,0 +1,223 @@
...
+	// during build time by setting this field.
+	// Ref: https://github.com/tinkerbell/hook/blob/main/linuxkit-templates/hook.template.yaml#L19
+	MagicString string
+}
+
+func (h *Handler) RoundTrip(req *http.Request) (*http.Response, error) {
+	h.Logger.V(1).Info("entered the roundtrip func")
+	if req.Method != http.MethodHead && req.Method != http.MethodGet {
+		return &http.Response{
+			Status:     "501 NOT IMPLEMENTED",
```
jacobweinstock, line 51:
```suggestion
			Status:     fmt.Sprintf("%v %s", http.StatusBadRequest, http.StatusText(http.StatusNotImplemented)),
```

jacobweinstock (changes requested):

jacobweinstock (approved):

